Issue #6, opened 2020-05-18
Separate object for a dataframe colum? (is Series needed?)


datapythonista:
Probably a bit early to the discussion, but I think this will need to be discussed eventually.

Is a separate object representing a single column needed? Like having Series, instead of just using one column DataFrame.

Having two separate objects, IMO adds a decent amount of complexity, both in the implementation and for the user. Whether this complexity is worth or not, I don't know. But I think this shouldn't be replicated from pandas without a discussion.

SimonHeybrock:
Sorry for being late to the party: My comment https://github.com/data-apis/dataframe-api/issues/59#issuecomment-913641267 is related to this. Similar xarray, [scipp](https://scipp.github.io/) supports what we can think of as a higher-dimensional generalization of `Series` (`DataArray`) with more explicit indices (coordinates). I think when working with higher-dimensional data then working with these `Series`-like objects is more common than working with a "dict-of-series" (`DataFrame`).

I also note some discussion about mask-columns above. Scipp does this explicitly by attaching not just a `coords` dict to an array of data values, but also a `masks` dict. See https://scipp.github.io/user-guide/masking.html for an introduction to the concept.

If there is interest I'd be happy to talk in more detail about these aspects.

rgommers:
It looks to me like there _may_ be a need for a separate column object - in the end that's what we needed in the interchange protocol as well - however the trick will then be to figure out how to add that without having a large set of duplicate APIs between column and dataframe objects.

jbrockmendel:
What would `df.iloc[n]` give in this scenario?  It might avoid casting in mixed-dtype cases which code be nice.

jbrockmendel:
> in what scenario?

```
df = pd.DataFrame({"A": [1], "B": [pd.Timestamp.now()]})
```

ATM `df.iloc[0]` gives an object-dtype Series.  IIUC in the proposal here `df.iloc[0]` would just return a view on `df`, so would not do any casting.

jbrockmendel:
Is there a .dtypes?  and if so, what does it return if there is no Series?

kkraus14:
> Is there a .dtypes? and if so, what does it return if there is no Series?

[...]

Issue #4, opened 2020-05-16
Avoiding the "pandas trap"


TomAugspurger:
Split from the discussions in https://github.com/pydata-apis/dataframe-api/issues/2.

To avoid the trap of "let's just match pandas", let's collect a list of specific problems with the pandas API, which we'll intentionally deviate from. To the extent possible we should limit this discussoin to issues with the API, rather than implementation.

---

- pandas.DataFrame can't implement [`collections.abc.Mapping`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping) because `.values` is a property returning an array, rather than a method. (added by @TomAugspurger)
- The "groupby-apply" pattern of passing opaque functions for non-trivial aggregations that are otherwise able to be expressed easily in e.g. SQL (consider: any aggregation expression that involves more than one column) (added by @wesm)
- Indexing in pandas accepts a variety of different inputs, which each have their own semantics, e.g. passing a function to the `__getitem__` or `loc`/`iloc`. It is not explicitly clear to new users the difference between `df[["a","b","c"]]` and `df[slice(5)]` and `df[lambda idx: idx % 5 == 0]`. (added by @devin-petersohn)
- pandas allows the dot operator (`__getattr__`) to get columns, which causes problems for columns that share names with other APIs. (added by @devin-petersohn)
- Duplicate APIs:
	- Simple aliases: e.g. `isna` and `isnull`, `multiply` and `mul`, etc.(added by @devin-petersohn)
	- More complex duplication: e.g. `query("a > b")` and `df[df["a"] > df["b"]]` (added by @devin-petersohn)
	- Indexing, there are 7 or 8 ways to get one or more columns in pandas: e.g. `__getitem__`, `__getattr__`, `loc`, `iloc`, `apply`, `drop` (added by @devin-petersohn)
	- `merge` and `join` call each other and are confusing for new users (added by @devin-petersohn)
- Having a separate object to represent a one column dataframe (i.e. `Series`). Creating all the complexity of having to reimplement most functionality of dataframe. And not providing a consistent way of applying operations to N columns (including 1). #6 @datapythonista 
- "missing" APIs / extension points:
[...]

PR #33, opened 2020-12-10
Add statistical methods

kgryte:
This PR

-   adds specifications for statistical methods.
-   is derived from comparing API [signatures](https://github.com/data-apis/dataframe-api-comparison/tree/2dae41fd0bdbd6b135470f640e0281f9315bc013/signatures/statistics) across dataframe libraries.

## Notes

-   Statistical methods are widely implemented across [dataframe libraries](https://github.com/data-apis/dataframe-api-comparison/blob/2dae41fd0bdbd6b135470f640e0281f9315bc013/data/common_apis.csv) and are used by downstream [libraries](https://github.com/data-apis/dataframe-api-comparison/blob/2dae41fd0bdbd6b135470f640e0281f9315bc013/data/common_apis_ranks.csv).

-   `Series` not included based on previous consortium discussions where dataframe/series distinction not considered necessary. See #4 and #6.

-   vaex and ibis have considerably different APIs than pandas, Dask, Modin, cuDF, and Koalas, and only influenced API inclusion based on whether the libraries provided a particular method name (or equivalent), but not keyword arguments.

-   Comments for each proposed method:

    -   **cummax**: only universally implemented keyword argument is `skipna`. Both cuDF and Koalas support `skipna`, but not `axis` (pandas, Dask, Modin).
    -   **cummin**: same comments as for `cummax`.
    -   **cumsum**: same comments as for `cummax`.
    -   **cumprod**: same comments as for `cummax`.
    -   **max**: only universally implemented keyword argument is `axis`. pandas, Modin, and Koalas support `numeric_only`, but others do not.
    -   **mean**: same comments as for `max`.
    -   **min**: same comments as for `max`.
    -   **nlargest**: no keywords arguments are universally implemented. cuDF does not support multiple column labels. This PR does specify that multiple labels be permitted. Not clear whether cuDF cannot technically support multiple column labels or whether this is not yet implemented.
    -   **nsmallest**: same comments as for `nlargest`.
    -   **prod**: same comments as for `max`. Koalas can only support positive numbers due to implementation algorithm. pandas, Dask, Modin, and cuDF support a `min_count` keyword argument, but Koalas does not.
    -   **std**: same comments as for `max`. Koalas does not support a correction factor. Similar to the array API specification, renamed `ddof` to `correction`, as this is a historical "bug" carried over from NumPy.
    -   **sum**: same comments as for `max`. pandas, Dask, Modin, and cuDF support a `min_count` keyword argument, but Koalas does not.
    -   **var**: same comments as for `std`.

-   methods excluded from this initial proposal: `mode`, `median`, `nunique`, and `quantile` due to either lack of universal availability, divergent behavior, increased complexity, or lack of downstream usage. These can be considered in a future PR.


rgommers:
As discussed, closing this PR given that we already have the most common reductions and the rest doesn't really fit with the direction we've taken over the last 6 months.

PR #35, opened 2021-01-05
Requirements document for the dataframe interchange protocol

rgommers:
Continued from gh-30. Main changes:
- added chunking
- added memory layout description including need to support multiple representations of missing data
- added some more detail on the Arrow C Data Interface
- addressed open review comments from gh-30

Also added one important TBD about the connection with the array data interchange protocol, with a suggestion to add that if and only if the supported dtypes allow (i.e., only boolean, integer and floating-point dtypes, and missing data as separate boolean masks).

Up-to-date rendered version: https://github.com/data-apis/dataframe-api/blob/dataframe-interchange-protocol/protocol/dataframe_protocol_summary.md


rgommers:
Hmm good question. The main motivation (from requirement 12 - maybe easier to read in rendered form in https://data-apis.org/dataframe-protocol/latest/design_requirements.html) is: _Must support chunking, i.e. accessing the data in “batches” of rows. _

Given that columns are contiguous in memory and rows are not, I don't think having different chunks on different columns is a problem, but there are some constraints probably. The reason for that is that both the `DataFrame` and the `Column` objects have `num_chunks`; if they're both used then I'd say a "dataframe chunk" is the outer container, and within that each column can be chunked. In case the dataframe has `num_chunks() == 1` then each column can in principle have an independent/arbitrary number of chunks.

Maybe a question to others - is there a reason to forbid this?

rgommers:
We had a discussion on this in a call 2 weeks ago, and the outcome was:

- Internally, it is possible for a dataframe to have different chunk sizes for different columns.
  - One example given was that it could have performance advantages in some cases when using a string column (this happens in cuDF it looks like).
- For interchange, uniform chunks are needed. The question is only if the re-chunking should be done on the producer or the consumer side.
- Preference was for this to be on the consumer side.
- Text and diagram here already show equal chunking, and at least the Modin implementation adhered to this (enforced on the producer side).
- What's needed is a doc clarification stating that the producer must produce equal-size chunks.

I'll open a follow-up PR.

PR #36, opened 2021-01-07
Add comparison methods

kgryte:
This PR

-   adds specifications for comparison methods.
-   is derived from comparing API [signatures](https://github.com/data-apis/dataframe-api-comparison/tree/296925cb9d88600414c56b7529fbf70e4c54f83e/signatures/comparison) across dataframe libraries.

## Notes

-   Comparison methods are widely implemented across [dataframe libraries](https://github.com/data-apis/dataframe-api-comparison/blob/2dae41fd0bdbd6b135470f640e0281f9315bc013/data/common_apis.csv) and are used by downstream [libraries](https://github.com/data-apis/dataframe-api-comparison/blob/2dae41fd0bdbd6b135470f640e0281f9315bc013/data/common_apis_ranks.csv).

-   For all comparison methods, cuDF hangs the methods on a `Series` object, but not the `DataFrame` object.

-   For all comparison methods, cuDF and Koalas do not support the `axis` keyword. The `axis` keyword is not currently included in this proposal (see note on broadcasting below).

-   For all comparison methods, cuDF and Koalas do not support the `level` keyword.

-   Pandas, Dask, Modin are liberal in what they accept (e.g., scalar, sequence, Series, DataFrame). This PR limits to DataFrame objects, similar to the array API specification. This choice should be discussed and confirmed.

-   As a follow-up to the prior note, this PR does not currently address broadcasting semantics. A few notes on this end:

    -   Similar to the array API specification, scalars can be broadcast as an MxN dataframe whose size matches the original dataframe. And similar to the array API specification, how libraries choose to represent a scalar as an MxN dataframe can be implementation-defined.
    -   A dataframe consisting of a single row (size: 1xN) can be broadcast as an MxN dataframe whose size matches the original dataframe.
    -   A dataframe consisting of a single column (size: Mx1) can be broadcast as an MxN dataframe whose size matches the original dataframe.
    -   Otherwise, the dataframe provided as `other` must have a size equal to the original dataframe (i.e., no broadcasting required).

### DIFF
--- a/spec/API_specification/comparison_methods.md
+++ b/spec/API_specification/comparison_methods.md
@@ -12,7 +12,7 @@ A conforming implementation of the dataframe API standard must provide and suppo
 <!-- NOTE: please keep the methods in alphabetical order -->
 
 (method-eq)=
-### eq(other, /, *, axis=None)
+### eq(other)
 
 Computes the truth value of `df_i == other_i` for each element `df_i` of the dataframe instance `df` with the respective element `other_i` of the dataframe `other`.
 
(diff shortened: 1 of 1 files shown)


rgommers:
This has been superceded by other PRs, I'll close this.

PR #37, opened 2021-01-07
Add math methods

kgryte:
This PR

-   adds specifications for math methods.
-   is derived from comparing API [signatures](https://github.com/data-apis/dataframe-api-comparison/tree/04fbc439285e440f105060a735bcc24bfd4e314b/signatures/math) across dataframe libraries.

## Notes

-   Math methods are widely implemented across [dataframe libraries](https://github.com/data-apis/dataframe-api-comparison/blob/2dae41fd0bdbd6b135470f640e0281f9315bc013/data/common_apis.csv) and are used by downstream [libraries](https://github.com/data-apis/dataframe-api-comparison/blob/2dae41fd0bdbd6b135470f640e0281f9315bc013/data/common_apis_ranks.csv).

-   For `abs`, cuDF provides support for `Series` but not `DataFrame` objects.

-   For binary (i.e., two-operand) operations, cuDF does not support `axis` or `level` keywords.

-   Koalas does not support `axis`, `level`, or `fill_value`.

-   The `axis` keyword is not currently included in this proposal. Whether this is desired depends on broadcasting semantics.

-   Pandas, Dask, Modin are liberal in what they accept (e.g., scalar, sequence, Series, DataFrame) for binary operations. This PR limits to DataFrame objects, similar to the array API specification. This choice should be discussed and confirmed.

-   This PR includes TODOs concerning broadcasting and type promotion. Semantics for both need to be confirmed, as may be linked to array semantics.

-   Open question as to how math methods should handle non-numerical columns/rows.

-   Open question as to supported data types? Arrays only handle integer and floating-point. Dataframes, such as `pandas`, support many more (e.g., complex, timedelta, etc).

-   Similar to array API, should dataframes be required to specify a default floating-point data type?

-   This PR does not make any assumptions regarding the underlying implementation (e.g., assumes array-API-specification-conforming arrays underneath the hood) and thus replicates floating-point specification behavior.

### DIFF
--- a/spec/API_specification/math_methods.md
+++ b/spec/API_specification/math_methods.md
@@ -1,0 +1,252 @@
+# Math Methods
+
+> Dataframe API specification for math methods.
+
+A conforming implementation of the dataframe API standard must provide and support the following methods adhering to the following conventions.
+
+-   Positional parameters must be [positional-only](https://www.python.org/dev/peps/pep-0570/) parameters. Positional-only parameters have no externally-usable name. When a method accepting positional-only parameters is called, positional arguments are mapped to these parameters based solely on their order.
+-   Optional parameters must be [keyword-only](https://www.python.org/dev/peps/pep-3102/) arguments.
+
+## Methods
+
+<!-- NOTE: please keep the methods in alphabetical order -->
+
+(method-abs)=
+### abs()
+
+Calculates the absolute value for each element of the dataframe instance `df` (i.e., the element-wise result has the same magnitude as the respective element in the original dataframe but has positive sign).
+
+#### Special Cases
+
+For floating-point operands,
+
+-   If an element is `NaN`, the result is `NaN`.
+-   If an element is `-0`, the result is `+0`.
+-   If an element is `-infinity`, the result is `+infinity`.
+
+#### Returns
+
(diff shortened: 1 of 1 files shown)


spec/API_specification/math_methods.md
```diff
@@ -0,0 +1,252 @@
...
+-   If an element is `-infinity`, the result is `+infinity`.
+
+#### Returns
+
+-   **out**: _&lt;dataframe&gt;_
+
+    -   dataframe containing element-wise results. Each element of the returned dataframe must have the same data type as the respective element in the original dataframe.
+
+(method-add)=
+### add(other, /)
```
jbrockmendel, line 34: pd.DataFrame ha both `add` and `__add__`.  This seems to refer to the former, whereas I think the latter would be a higher priority

rgommers: There should probably be a separate section about Python operator support, ala https://data-apis.org/array-api/latest/API_specification/array_object.html#operators. And a design rule that for every operator that the 3 flavors (e.g., `+`, `__add__` and `add`) must have matching behavior.

rgommers:
This has been superceded by other PRs, I'll close this.

PR #43, opened 2021-06-23
PR: Add metadata attribute to DataFrame and Column

steff456:
This PR adds a metadata attribute that can be used to store library-specific things. For example, libraries like Vaex will  be able to store expressions for its virtual columns there.

- [x] Add metadata attribute to the DataFrame class
- [x] Add metadata attribute to the Column class
- [x] Add a test


rgommers (approved):
Thanks @steff456, merged in commit d9419b1.

PR #45, opened 2021-06-24
Add variable-length string support

kgryte:
This PR

-   adds variable-length string support to the dataframe interchange protocol.
-   modifies the protocol to return a dictionary containing all possible data buffers, rather than separate methods for retrieving data, validity, and offsets buffers, as discussed in consortium meetings.
-   modifies the protocol to return both the buffers and their associated dtypes. This is necessary in order to interpret, e.g., the `offsets` and `mask` buffers.
-   updates source code documentation (punctuation, typos, and spelling).
-   manually copies string data to and from byte arrays, as it assumes that strings are stored as `object` dtype. The implementation will need to be updated to accommodate pandas' string extension dtype which is based on arrow. Currently, the string extension dtype is considered experimental and subject to change. The use of `object` dtype is still used as the default string dtype for backward compatiblity.
-    uses a byte array mask for indicating missing values in string data buffers. This can be updated to use a bit array for space efficiency, at the cost of additional code complexity.
-    currently does not handle the scenario where a non-pandas dataframe provided to `__dataframe__` uses a bit array to indicate missing values.


rgommers:
Thanks @kgryte! It may be useful to close this PR and resend it against `main`, now that the other PRs it needed are merged.

protocol/pandas_implementation.py
```diff
@@ -160,11 +162,55 @@ def convert_categorical_column(col : ColumnObject) -> pd.Series:
...
+    dbuf = buffer_to_ndarray(dbuffer, dt)
+
+    obuf = buffer_to_ndarray(obuffer, odtype)
+    mbuf = buffer_to_ndarray(mbuffer, mdtype)
+
+    # Assemble the strings from the code units
+    str_list = []
+    for i in range(obuf.size-1):
+        # Check for missing values
+        if mbuf[i] == 0:  # FIXME: we need to account for a mask buffer which is a bit array
```
rgommers, line 189: The mask buffer can have multiple formats, see `describe_null`.

protocol/pandas_implementation.py
```diff
@@ -452,27 +507,97 @@ def get_data_buffer(self) -> Tuple[_PandasBuffer, Any]:  # Any is for self.dtype
             codes = self._col.values.codes
             buffer = _PandasBuffer(codes)
             dtype = self._dtype_from_pandasdtype(codes.dtype)
+        elif self.dtype[0] == _k.STRING:
+            # Marshal the strings from a NumPy object array into a byte array
```
rgommers, line 511: To preempt some complaints: maybe add a comment like "yes this for-loop is slow, it can be reimplemented in Cython/C/C++ later"

protocol/pandas_implementation.py
```diff
@@ -452,27 +507,97 @@ def get_data_buffer(self) -> Tuple[_PandasBuffer, Any]:  # Any is for self.dtype
...
         """
+        _k = _DtypeKind
+        if self.dtype[0] == _k.STRING:
+            # For now, have the mask array be comprised of bytes, rather than a bit array
+            buf = self._col.to_numpy()
+            mask = []
+            for i in range(buf.size):
+                v = 0;
+                if type(buf[i]) == str:
+                    v += 1;  # follows Arrow where a valid value is 1 and null is 0
```
rgommers, line 543: Argh, we didn't specify this yet. The name `mask` suggests the opposite to me (as does `numpy.ma`):
```python
>>> x = np.ma.arange(5)
>>> x.mask = [0, 1, 0, 1, 0]
>>> x
masked_array(data=[0, --, 2, --, 4],
             mask=[False,  True, False,  True, False],
       fill_value=999999)
```

rgommers: maybe it should be called `validity` for clarity:)

rgommers (commented):
This looks very good, thanks @kgryte! I haven't tested it yet, but just reading through the code I have only a couple of comments.

protocol/pandas_implementation.py
kgryte: Aware that `null` can take on one of different values (ala `describe_null`); however, the mask buffer can only be a byte array or bit array, correct? Meaning, we only need to account for one of two scenarios. What remains to be decided, however, is what value a missing value should take on (`1` or `0`; Arrow vs NumPy).

protocol/pandas_implementation.py
kgryte: Added.

protocol/pandas_implementation.py
kgryte: Yeah, noticed that Arrow does the opposite of NumPy. We should probably add this to the dataframe meeting agenda.

kgryte:
@rgommers Will close this and submit against `main` in short order.

protocol/pandas_implementation.py
kgryte: For purposes of documenting the divergence in this PR...

From Arrow [docs](https://arrow.apache.org/docs/format/Columnar.html#validity-bitmaps):

> A 1 (set bit) for index j indicates that the value is not null, while a 0 (bit not set) indicates that it is null.

From NumPy [docs](https://numpy.org/doc/stable/reference/maskedarray.generic.html#what-is-a-masked-array):

> When an element of the mask is False, the corresponding element of the associated array is valid and is said to be unmasked. When an element of the mask is True, the corresponding element of the associated array is said to be masked (invalid).

jorisvandenbossche:
> Will close this and submit against main in short order.

You can actually change the target branch by clicking the "Edit" button next to title, so then you don't need to close / open a new PR

protocol/pandas_implementation.py
jorisvandenbossche: > What remains to be decided, however, is what value a missing value should take on (1 or 0; Arrow vs NumPy).

That can depend on byte vs bit? (byte -> follow numpy, bit -> follow Arrow)

protocol/pandas_implementation.py
jorisvandenbossche: > maybe it should be called validity for clarity:)

That's how it is called in Arrow :) (although the C++ implementation typically uses null_bitmap)

Instead of `get_mask()`, it could also be something more generic like `get_null_buffer()` (in line of `get_data_buffer`)

protocol/dataframe_protocol_summary.md
```diff
@@ -209,14 +220,16 @@ except `__dataframe__` is a Python-level rather than C-level interface.
 The data types format specification of that interface is something that could
 be used unchanged.
 
-The main (only?) limitation seems to be that it does not have device support
-- @kkraus14 will bring this up on the Arrow dev mailing list. Also note that
-that interface only talks about arrays; dataframes, chunking and the metadata
-inspection can all be layered on top in this Python-level protocol, but are
-not discussed in the interface itself.
+The main limitation is to be that it does not have device support
```
jorisvandenbossche, line 223:
```suggestion
The main limitation is that it does not have device support
```

protocol/pandas_implementation.py
```diff
@@ -452,27 +507,99 @@ def get_data_buffer(self) -> Tuple[_PandasBuffer, Any]:  # Any is for self.dtype
...
         """
-        Return the buffer containing the mask values indicating missing data.
+        Return the buffer containing the mask values indicating missing data and
+        the buffer's associated dtype.
 
         Raises RuntimeError if null representation is not a bit or byte mask.
         """
+        _k = _DtypeKind
+        if self.dtype[0] == _k.STRING:
+            # For now, have the mask array be comprised of bytes, rather than a bit array
```
jorisvandenbossche, line 539: Shouldn't this actually raise a RuntimeError, since pandas uses the NaN sentinel for strings? In which case there should be no null mask. `describe_null` above is also updated to return 1 (NaN approach)

jorisvandenbossche: Of course, although pandas is using that (currently), that doesn't work with the actual return data representation here for the string data (one string buffer with offsets). But so then `describe_null` should be updated to indicate that it uses a byte array approach.

protocol/pandas_implementation.py
```diff
@@ -452,27 +507,99 @@ def get_data_buffer(self) -> Tuple[_PandasBuffer, Any]:  # Any is for self.dtype
...
+            for v in values:
+                # For missing values (in this case, `np.nan` values), we don't increment the pointer)
+                if type(v) == str:
+                    b = v.encode(encoding="utf-8")
+                    ptr += len(b)
+
+                offsets.append(ptr)
+
+            # Convert the list of offsets to a NumPy array of signed 64-bit integers (note: Arrow allows the offsets array to be either `int32` or `int64`; here, we default to the latter)
+            buf = np.asarray(offsets, dtype="int64")
```
jorisvandenbossche, line 590: Do we want to specify this? (the int32 vs int64) Or is the implementer free to use one or the other? 

(in Arrow this is a different type, although that is not always convenient)

jorisvandenbossche (commented):
I think the separate `get_data_buffer` and `get_offsets` methods might be a bit problematic. Typically you only want to do one pass over the data and create all buffers at once. 

I know this is only a dummy implementation, and presumably the created buffers could be cached on the object so the separate methods don't calculate it twice. But might be useful to think about separate methods vs a single `get_buffers()` (with a specified order)

protocol/pandas_implementation.py
kgryte: Would it not be confusing to follow two different (and opposite) conventions? I would think it better to be consistent.

kgryte:
> You can actually change the target branch by clicking the "Edit" button next to title, so then you don't need to close / open a new PR

I initially tried doing that, but doing so added many unrelated changes and muddied this PR. :(

protocol/pandas_implementation.py
kgryte: Yeah, I found this tricky. I was not certain which reality `describe_null` should describe. Currently, the implementation both uses `NaN` as a sentinel value (following pandas), but also needs the mask buffer for interchange, as the data buffer is only a string buffer (byte array).

I think it would make sense to update `describe_null` as you describe while leaving the column free to use `NaN` for those values in the mask array which are invalid.

protocol/pandas_implementation.py
kgryte: Renamed to `get_validity_buffer`.

protocol/pandas_implementation.py
kgryte: I've updated `describe_null` to indicate that the protocol uses a byte array `(4, None)` for the string dtype.

protocol/pandas_implementation.py
kgryte: My sense is that the implementer should be free to choose. When accessing the `offsets` buffer, we also return the buffer's associated dtype info, so consumers should be able to determine how they should interpret the returned buffer.

jorisvandenbossche:
> I initially tried doing that, but doing so added many unrelated changes and muddied this PR. :(

I _think_ that either merging latest master in this branch or rebasing this branch on top of master should solve that

kgryte:
Closing this PR out in favor of [gh-47](https://github.com/data-apis/dataframe-api/pull/47).

PR #47, opened 2021-07-19
Add variable-length string support

kgryte:
This PR supersedes [gh-45](https://github.com/data-apis/dataframe-api/pull/45) in order to ensure a cleaner merge and, similar to [gh-45](https://github.com/data-apis/dataframe-api/pull/45),

-   adds variable-length string support to the dataframe interchange protocol.
-   modifies the protocol to return a dictionary containing all possible data buffers, rather than separate methods for retrieving data, validity, and offsets buffers, as discussed in consortium meetings.
-   modifies the protocol to return both the buffers and their associated dtypes. This is necessary in order to interpret, e.g., the `offsets` and `mask` buffers.
-   updates source code documentation (punctuation, typos, and spelling).
-   manually copies string data to and from byte arrays, as it assumes that strings are stored as `object` dtype. The implementation will need to be updated to accommodate pandas' string extension dtype which is based on arrow. At the time of this PR, the string extension dtype was considered experimental and subject to change. The use of `object` dtype is still used as the default string dtype for backward compatiblity.
-    uses a byte array mask for indicating missing values in string data buffers. This can be updated to use a bit array for space efficiency, at the cost of additional code complexity.

### DIFF
--- a/protocol/dataframe_protocol.py
+++ b/protocol/dataframe_protocol.py
@@ -308,20 +308,20 @@ class Column:
 
         The returned dictionary has the following contents:
 
-            - "data": a two-element tuple whose first element is a tuple
-                      containing the buffer containing the data and whose second
-                      element is the data buffer's associated dtype.
-            - "validity": a two-element tuple whose first element is a tuple
-                          containing the buffer containing mask values
-                          indicating missing data and whose second element is
-                          the mask value buffer's associated dtype. None if the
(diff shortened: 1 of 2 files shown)


rgommers (approved):
This LGTM, I played with it a little and it all seems to work as expected. The main review comments are on gh-45, and those were all addressed. It's time (well, overdue) to merge this. Thanks @kgryte and all reviewers!

Issue #46, opened 2021-06-25
Tracking issue: dataframe protocol implementation


rgommers:
The bulk of the dataframe interchange protocol was done in gh-38. There were still a number of TODOs however, and more will likely pop up once we have multiple implementations so we can actually turn one type of dataframe into another type. This is the tracking issue for those TODOs and issues:

- [ ] Categorical dtypes: we should allow having `null` as a category; it should not have a specified meaning, it's just another category that should (e.g.) roundtrip correctly. See conversation in 8 Apr meeting.
- [ ] Categorical dtypes: should they be a dtype in themselves, or should they be a part of the dtype tuple? Currently dtype is `(kind, bitwidth, format_str, endianness)`, with categorical being a value of the `kind` enum. Is making a 5th element in the dtype, with that element being another dtype 4-tuple, thereby allowing for nesting, sensible?
- [x] Add a `metadata` attribute that can be used to store library-specific things. For example, Vaex should be able to store expressions for its virtual columns there. _See PR gh-43_
- [x] Add a flag to throw an exception if the export cannot be zero-copy. (e.g. for pandas, possible due to block manager where rows are contiguous and columns are not - add a test for that). _See PR gh-44_
- [x] Add a string dtype, with variable-length strings implemented with the same scheme as Arrow uses (an `offsets` and a `data` buffer, see https://github.com/data-apis/dataframe-api/pull/38#discussion_r609818874). _See PR gh-45
- [x] Signature of the `from_dataframe` protocol? See https://github.com/data-apis/dataframe-api/issues/42 and meeting of 20 May.
- [x] What can be reused between implementations in different libraries, and can/should we have a reference implementation? --> question needs answering somewhere.
- [ ] What is the ownership for buffers, who owns the memory? This should be clearly spelled out in the docs. An `owner` attribute is perhaps needed. See meeting minutes 4 March, https://github.com/data-apis/dataframe-api/issues/39, and comments on this PR.

PR #49, opened 2021-08-17
More tests on the API

iskode:
Current tests focus on `from_dataframe` method. To have a better test coverage, I've written some test cases for `__dataframe__` method. Note that this is a starting point. Added tests are not complete e.g `Buffer` tests are missing.

### DIFF
--- a/protocol/pandas_implementation.py
+++ b/protocol/pandas_implementation.py
@@ -538,23 +538,25 @@ class _PandasDataFrame:
 # Roundtrip testing
 # -----------------
 
-def test_column(pdcol:pd.Series, col: _PandasColumn):
-    assert pdcol.size == col.size
+def assert_column_equal(col: _PandasColumn, pdcol:pd.Series):
+    assert col.size == pdcol.size 
     assert col.offset == 0
-    assert pdcol.isnull().sum() == col.null_count 
+    assert col.null_count == pdcol.isnull().sum() 
+    assert col.num_chunks() == 1
+    pytest.raises(RuntimeError, col.get_mask)
 
-def test__dataframe__(df:pd.DataFrame, dfo: DataFrameObject):
+def assert_dataframe_equal(dfo: DataFrameObject, df:pd.DataFrame):
     assert dfo.num_columns() == len(df.columns)
     assert dfo.num_rows() == len(df)
     assert dfo.num_chunks() == 1
     assert dfo.column_names() == list(df.columns)
     for col in df.columns:
-        test_column(df[col], dfo.get_column_by_name(col))
+        assert_column_equal(dfo.get_column_by_name(col), df[col])
 
 def test_float_only():
     df = pd.DataFrame(data=dict(a=[1.5, 2.5, 3.5], b=[9.2, 10.5, 11.8]))
     df2 = from_dataframe(df)
-    test__dataframe__(df, df.__dataframe__())
+    assert_dataframe_equal(df.__dataframe__(), df)
     tm.assert_frame_equal(df, df2)
 
 
@@ -562,7 +564,7 @@ def test_mixed_intfloat():
     df = pd.DataFrame(data=dict(a=[1, 2, 3], b=[3, 4, 5],
                                 c=[1.5, 2.5, 3.5], d=[9, 10, 11]))
     df2 = from_dataframe(df)
-    test__dataframe__(df, df.__dataframe__())
+    assert_dataframe_equal(df.__dataframe__(), df)
     tm.assert_frame_equal(df, df2)
 
 
@@ -590,7 +592,7 @@ def test_categorical_dtype():
     assert col.describe_categorical == (False, True, {0: 1, 1: 2, 2: 5})
 
     df2 = from_dataframe(df)
-    test__dataframe__(df, df.__dataframe__())
+    assert_dataframe_equal(df.__dataframe__(), df)
     tm.assert_frame_equal(df, df2)
 
 
--- a/protocol/pandas_implementation.py
+++ b/protocol/pandas_implementation.py
@@ -538,12 +538,29 @@ class _PandasDataFrame:
 # Roundtrip testing
 # -----------------
 
+def assert_buffer_equal(buffer_dtype: Tuple[_PandasBuffer, Any], pdcol:pd.Series):
+    buf, dtype = buffer_dtype
+    pytest.raises(NotImplementedError, buf.__dlpack__)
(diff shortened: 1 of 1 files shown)


protocol/pandas_implementation.py
```diff
@@ -538,16 +538,31 @@ def get_chunks(self, n_chunks : Optional[int] = None) -> Iterable['_PandasDataFr
 # Roundtrip testing
 # -----------------
 
+def test_column(pdcol:pd.Series, col: _PandasColumn):
```
kshitij12345, line 541: Since this looks to be a helper function used only by `test__dataframe__`, why not make it a nested function within `test_dataframe__`?

protocol/pandas_implementation.py
```diff
@@ -538,16 +538,31 @@ def get_chunks(self, n_chunks : Optional[int] = None) -> Iterable['_PandasDataFr
...
+    assert pdcol.size == col.size
+    assert col.offset == 0
+    assert pdcol.isnull().sum() == col.null_count 
+
+def test__dataframe__(df:pd.DataFrame, dfo: DataFrameObject):
```
kshitij12345, line 546: Should this be renamed to `_test_dataframe_protocol` (note the leading underscore since this looks to be a test helper and not an actual test)? `test__dataframe__` looks a bit odd.

kshitij12345 (commented):
Made a couple of minor comments. Would love to know your thoughts.

Thanks! @iskode

protocol/pandas_implementation.py
rgommers: May make sense. It could also be needed separately later, so top-level may make sense. But it's not a test, so name shouldn't start with `test`.

Often these kinds of functions are called something like `assert_column_equal`.

protocol/pandas_implementation.py
rgommers: Same suggestion as for the column function above: maybe something like `assert_dataframe_equal` (or `assert_df_equal`, or ...)?

rgommers (commented):
Thanks @iskode. Modulo naming things, I think this test approach does make sense. The `assert_frame_equal` checks are more comprehensive, however a separate check that the structure of the object created by `__dataframe__()` is correct is useful.

protocol/pandas_implementation.py
iskode: Thank you for your feedback. I'll add your suggestions.

protocol/pandas_implementation.py
iskode: Thank you again. I'll add this suggestion.

iskode:
I've found some issues with testing the buffer in case of categorical variables. Commented codes in the function `assert_buffer_equal` don't pass.

protocol/pandas_implementation.py
```diff
@@ -538,16 +538,50 @@ def get_chunks(self, n_chunks : Optional[int] = None) -> Iterable['_PandasDataFr
...
+    # assert dtype[1] == col.dtype.itemsize * 8, f"{dtype[1]} is not {col.dtype.itemsize * 8}"
+    # assert dtype[2] == col.dtype.str, f"{dtype[2]} is not {col.dtype.str}"
+
+
+def assert_column_equal(col: _PandasColumn, pdcol:pd.Series):
+    assert col.size == pdcol.size 
+    assert col.offset == 0
+    assert col.null_count == pdcol.isnull().sum() 
+    assert col.num_chunks() == 1
+    pytest.raises(RuntimeError, col.get_mask)
```
rgommers, line 562: Note that this method is renamed - there's a single `get_buffers()` call now which returns a dictionary. This line seems incorrect to me though, retrieving the mask should not always raise right?

protocol/pandas_implementation.py
```diff
@@ -538,16 +538,50 @@ def get_chunks(self, n_chunks : Optional[int] = None) -> Iterable['_PandasDataFr
...
+    assert_buffer_equal(col.get_data_buffer(), pdcol)
```
rgommers, line 563: `col._get_data_buffer` or `col.get_buffers()['data']`

protocol/pandas_implementation.py
```diff
@@ -538,16 +538,50 @@ def get_chunks(self, n_chunks : Optional[int] = None) -> Iterable['_PandasDataFr
 # Roundtrip testing
 # -----------------
 
+def assert_buffer_equal(buffer_dtype: Tuple[_PandasBuffer, Any], pdcol:pd.Series):
+    buf, dtype = buffer_dtype
+    pytest.raises(NotImplementedError, buf.__dlpack__)
+    assert buf.__dlpack_device__() == (1, None)
+    # It seems that `bitwidth` is handled differently for `int` and `category`
+    # assert dtype[1] == pdcol.dtype.itemsize * 8, f"{dtype[1]} is not {pdcol.dtype.itemsize}"
```
rgommers, line 546: Yes, something is wrong here. The root cause is in `_get_data_buffer`:
```python
        elif self.dtype[0] == _k.CATEGORICAL:
            buffer = _PandasBuffer(
                codes, allow_copy=self._allow_copy)
            dtype = self._dtype_from_pandasdtype(codes.dtype)
```
The categorical column has `codes` which are `int8` and categories which are `int64`. The series dtype shows the categories while the buffer dtype uses the codes:
```python
In [7]: pdcol.dtype
Out[7]: CategoricalDtype(categories=[1, 2, 5], ordered=False)

In [8]: pdcol.dtype.itemsize
Out[8]: 8

In [9]: dtype
Out[9]: (<_DtypeKind.INT: 0>, 8, '|i1', '|'
```

rgommers: Categorical data representation is incomplete; dtype support is there but the buffer support isn't. Also needs more tests.

protocol/pandas_implementation.py
rgommers: (this came in when the variable-length string support was merged.

protocol/pandas_implementation.py
rgommers: I'd suggest leaving it commented out for now. The rest of this PR looks good.

rgommers: See the first two items in the description of gh-46 for thoughts on what changes are still needed for categorical.

rgommers: @AlenkaF this test case and comment above about gh-46 is what you were mentioning as well about representation of categorical dtypes.

rgommers (approved):
This LGTM now, thanks @iskode

PR #51, opened 2021-08-23
Add `allow_copy` flag to interchange protocol

rgommers:
Rebase and fix up of gh-44.

### DIFF
--- a/protocol/dataframe_protocol.py
+++ b/protocol/dataframe_protocol.py
@@ -347,7 +347,8 @@ class DataFrame:
     ``__dataframe__`` method of a public data frame class in a library adhering
     to the dataframe interchange protocol specification.
     """
-    def __dataframe__(self, nan_as_null : bool = False) -> dict:
+    def __dataframe__(self, nan_as_null : bool = False,
+                      allow_copy : bool = True) -> dict:
         """
         Produces a dictionary object following the dataframe protocol specification.
 
@@ -355,8 +356,14 @@ class DataFrame:
         producer to overwrite null values in the data with ``NaN`` (or ``NaT``).
         It is intended for cases where the consumer does not support the bit
         mask or byte mask that is the producer's native representation.
+
+        ``allow_copy`` is a keyword that defines whether or not the library is
+        allowed to make a copy of the data. This can for example happen if a
+        library supports strided buffers (those would need a copy, because this
+        protocol specifies contiguous buffers).
         """
         self._nan_as_null = nan_as_null
+        self._allow_zero_zopy = allow_copy
         return {
             "dataframe": self,  # DataFrame object adhering to the protocol
             "version": 0        # Version number of the protocol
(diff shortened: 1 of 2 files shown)


protocol/dataframe_protocol.py
```diff
@@ -347,16 +347,23 @@ class DataFrame:
...
 
         ``nan_as_null`` is a keyword intended for the consumer to tell the
         producer to overwrite null values in the data with ``NaN`` (or ``NaT``).
         It is intended for cases where the consumer does not support the bit
         mask or byte mask that is the producer's native representation.
+
+        ``allow_copy`` is a keyword that defines whether or not the library is
+        allowed to make a copy of the data. This can for example happen if a
+        library supports strided buffers (those would need a copy, because this
+        protocol specifies contiguous buffers).
```
kgryte, line 363:
```suggestion
        allowed to make a copy of the data. For example, copying data would be
        necessary if a library supports strided buffers, given that this protocol
        specifies contiguous buffers.
```

protocol/pandas_implementation.py
```diff
@@ -232,12 +239,21 @@ def __dataframe__(cls, nan_as_null : bool = False) -> dict:
     producer to overwrite null values in the data with ``NaN`` (or ``NaT``).
     This currently has no effect; once support for nullable extension
     dtypes is added, this value should be propagated to columns.
+
+    ``allow_copy`` is a keyword that defines whether or not the library is
+    allowed to make a copy of the data. This can for example happen if a
+    library supports strided buffers (those would need a copy, because this
+    protocol specifies contiguous buffers).
```
kgryte, line 246:
```suggestion
    allowed to make a copy of the data. For example, copying data would be
    necessary if a library supports strided buffers, given that this protocol
    specifies contiguous buffers.
```

protocol/pandas_implementation.py
```diff
@@ -553,11 +573,13 @@ def _get_data_buffer(self) -> Tuple[_PandasBuffer, Any]:  # Any is for self.dtyp
...
         _k = _DtypeKind
         if self.dtype[0] in (_k.INT, _k.UINT, _k.FLOAT, _k.BOOL):
-            buffer = _PandasBuffer(self._col.to_numpy())
+            buffer = _PandasBuffer(
+                self._col.to_numpy(), allow_copy=self._allow_copy)
             dtype = self.dtype
         elif self.dtype[0] == _k.CATEGORICAL:
             codes = self._col.values.codes
-            buffer = _PandasBuffer(codes)
+            buffer = _PandasBuffer(
```
kgryte, line 581: As a sanity check, we're not similarly passing `allow_copy` at L595, L634, and L676 because we have guaranteed contiguous buffers in those cases?

kgryte (approved):
LGTM. Nothing obvious jumps out at me. One minor clarification question left in the review regarding when we're choosing to pass `allow_copy`.

protocol/pandas_implementation.py
rgommers: Yes indeed. `np.asarray(some_list)` creates a new contiguous array, and the `bytearray` usage seems to be contiguous too (that one was a bit harder to verify, but I did test it).

rgommers:
Okay, in it goes then. Thanks for verifying @kgryte

Issue #54, opened 2021-08-25
Arrow format string for format_str in _dtype_from_vaexdtype()


AlenkaF:
The code now uses NumPy format strings, while the docs for Column.dtype specify it must use the format string from the Apache Arrow C Data Interface (similar but slightly different). So we need a utility to map NumPy to Arrow format here. 

Example - should say 'b' not |b1':

``` python
df = pd.DataFrame({"A": [True, False, False, True]})

>>> df.__dataframe__().get_column_by_name('A').dtype
(<_DtypeKind.BOOL: 20>, 8, '|b1', '|')
```

Source:
https://arrow.apache.org/docs/format/CDataInterface.html#data-type-description-format-strings
https://numpy.org/doc/stable/reference/arrays.interface.html#arrays-interface
https://numpy.org/doc/stable/reference/generated/numpy.dtype.itemsize.html
https://numpy.org/doc/stable/reference/generated/numpy.dtype.byteorder.html

rgommers:
@honno could you please verify that https://github.com/data-apis/dataframe-interchange-tests has a test checking that the format strings used in each dataframe library are indeed in Arrow format now?

xref https://github.com/data-apis/dataframe-api/pull/62, which has a basic test that may be reusable in case you don't yet have one.

honno:
> could you please verify that https://github.com/data-apis/dataframe-interchange-tests has a test checking that the format strings used in each dataframe library are indeed in Arrow format now?

I had left it as a TODO, but just now updated [`test_dtype`](https://github.com/data-apis/dataframe-interchange-tests/blob/73e4f74eb53eb04bc9632ea243d8b167c458dbfe/tests/test_column_object.py#L88) to test that the format strings are in Arrow format (for the most part—will explore testing some trickier scenarios later). I found vaex and cuDF were returning the NumPy-style format strings and filed issues respectively.

rgommers:
Thanks @honno. I think we are good here then. I double checked for Pandas, and that looks as expected:
```python
>>> import pandas as pd
>>> pd.__version__
'1.5.0rc0'
>>> df = pd.DataFrame({"A": [True, False, False, True]})
>>> df.__dataframe__().get_column_by_name('A').dtype
(<DtypeKind.BOOL: 20>, 8, 'b', '|')
```

Let's close this. Thanks @AlenkaF and @honno

jorisvandenbossche:
[...]

PR #62, opened 2021-09-07
Add _format_str mapping to interchange protocol

AlenkaF:
This PR is a proposal to fix issue https://github.com/data-apis/dataframe-api/issues/54:
method `_format_str` was added to `_PandasColumn` with mapping of basic string formats from NumPy to Apache Arrow C Data Interface.

cc @kshitij12345 @rgommers

### DIFF
--- a/protocol/pandas_implementation.py
+++ b/protocol/pandas_implementation.py
@@ -902,6 +902,12 @@ def test_metadata():
     assert_dataframe_equal(df.__dataframe__(), df)
     tm.assert_frame_equal(df, df2)
 
+def test_fromat_str():
+    df = pd.DataFrame(data=dict(a=[1, 2, 3], B=[3, 4, 5],
+                                c=[1.5, 2.5, 3.5], D=["a", "b", "cdef"]))
+    df["B"] = df["B"].astype("category")
+    df["D"] = df["D"].astype("object")
+    df2 = from_dataframe(df)
 
 if __name__ == '__main__':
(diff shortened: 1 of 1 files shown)


protocol/pandas_implementation.py
```diff
@@ -435,10 +435,32 @@ def _dtype_from_pandasdtype(self, dtype) -> Tuple[enum.IntEnum, int, str, str]:
...
 
         bitwidth = dtype.itemsize * 8
-        format_str = dtype.str
+        format_str = self._format_str(dtype.str)
         endianness = dtype.byteorder if not kind == _k.CATEGORICAL else '='
         return (kind, bitwidth, format_str, endianness)
 
+    def _format_str(self, format_str) -> str:
+        """
+        Mapping of NumPy formt strings to
```
kgryte, line 444:
```suggestion
        Mapping of NumPy format strings to
```

rgommers:
Thanks a lot @AlenkaF. Given that after gh-81 the implementation has been removed from this repo, and Pandas now has its own implementation merged (with significant improvements compared to the prototype that this PR was refactoring), I will close this PR. 

We did learn quite a bit from your Vaex implementation work. Thanks for opening this PR, and apologies it did not get merged last year.

I will leave gh-54 open until we've verified that there's a test for this in https://github.com/data-apis/dataframe-interchange-tests, so we ensure that all four known implementations are indeed using the Arrow format strings.

Issue #41, opened 2021-04-08
storing & exchange of categorical dtypes


rgommers:
# Categorical dtypes

xref gh-26 for some discussion on categorical dtypes.

### Pandas

The dtype is called `category` there. See [pandas.Categorical docs](https://pandas.pydata.org/docs/reference/api/pandas.Categorical.html):

```python
>>> df = pd.DataFrame({"A": [1, 2, 5, 1]})
>>> df["B"] = df["A"].astype("category")

>>> df.dtypes
A       int64
B    category
dtype: object

>>> col = df['B']
>>> col.dtype
CategoricalDtype(categories=[1, 2, 5], ordered=False)

>>> col.values.ordered
False
>>> col.values.codes
array([0, 1, 2, 0], dtype=int8)
>>> col.values.categories
Int64Index([1, 2, 5], dtype='int64')
>>> col.values.categories.values
array([1, 2, 5])
```

### Apache Arrow

The dtype is called _"dictionary-encoded" in Arrow - so a dataframe with a categorical dtype is called a "dictionary-encoded array" there.
See https://arrow.apache.org/docs/format/CDataInterface.html#structure-definitions for details.

A practical example (from @kkraus14 in gh-38), for a categorical column of
`['gold', 'bronze', 'silver', null, 'bronze', 'silver', 'gold']` with categories of
`['gold' < 'silver' < 'bronze']`:
```
categorical column: {
    mask_buffer: [119], # 01110111 in binary
    data_buffer: [0, 2, 1, 127, 2, 1, 0], # the 127 value in here is undefined since it's null
    children: [
        string column: {
            mask_buffer: None,
            offsets_buffer: [0, 4, 10, 16],
            data_buffer: [103, 111, 108, 100, 115, 105, 108, 118, 101, 114, 98, 114, 111, 110, 122, 101]
        }
    ]
}
```

```C++
struct ArrowSchema {
  // Array type description
  const char* format;
  const char* name;
  const char* metadata;
  int64_t flags;
  int64_t n_children;
  struct ArrowSchema** children;
  struct ArrowSchema* dictionary;  // the categories
  ...
};

struct ArrowArray {
  // Array data description
  int64_t length;
  int64_t null_count;
  int64_t offset;
  int64_t n_buffers;
  int64_t n_children;
  const void** buffers;
  struct ArrowArray** children;
  struct ArrowArray* dictionary;
  ...
};
```

[...]

PR #69, opened 2021-09-29
Use a column to store categories, rather than a mapping

shwina:
This PR implements one of the changes mentioned in https://github.com/data-apis/dataframe-api/issues/41, i.e.,

> Add get_children() method, and store the mapping that is now in Column.describe_categorical in a child column instead. Note that child columns are also needed for variable-length strings.

Replacing the previous `mapping` with a child column now implies that the data buffer stores integer indices into the child column.


kkraus14 (approved):

vnlitvinov:
@shwina @rgommers should we maybe not try to disguise the (explicit) categories as generic "children"?

I think just replacing the _type_ of `df.describe_categorical['mapping']` from `dict` to `Column` instead of introducing of a new method could be better - it's still a mapping after all 🙃

rgommers:
We briefly touched on this PR today, we'd like to move it forward since all interchange protocol implementers seem to be in favor of this. @vnlitvinov plans to revisit his proposal from above and add that here or open an alternative PR. Then we can try to finalize this, and then update implementations in the various libraries for it.

shwina:
> still a mapping after all

I'm not sure I quite follow; the `Column` is explicitly _not_ a mapping, rather it is concretely the `categories`. Perhaps I'm missing something?

vnlitvinov:
@shwina I've missed your comment, so I went ahead with the original plan we discussed last meeting and I had added a commit doing so. If you don't like the idea feel free to remove my commit.

As for it not being a mapping... it somewhat is :) it maps an integer index to a category value.

rgommers:
After discussion on Thursday, @shwina and others were happy with the changes @vnlitvinov pushed. With one change to make: rename `mapping` to `categories`, since it's no longer a real mapping (only implicitly, the keys are `0 ... n_categories`) and it would help with transitioning too.

protocol/dataframe_protocol.py
```diff
@@ -170,23 +170,24 @@ def dtype(self) -> Tuple[enum.IntEnum, int, str, str]:
...
 
     @property
-    def describe_categorical(self) -> dict[bool, bool, Optional[dict]]:
+    def describe_categorical(self) -> dict[bool, bool, Optional[Column]]:
         """
         If the dtype is categorical, there are two options:
 
         - There are only values in the data buffer.
-        - There is a separate dictionary-style encoding for categorical values.
+        - There is a separate non-categortical Column encoding categorical values.
```
kkraus14, line 178:
```suggestion
        - There is a separate non-categorical Column encoding categorical values.
```

vnlitvinov: nice catch, thanks!

rgommers:
> With one change to make: rename `mapping` to `categories`,

This change was made in the spec part, but not the rest of this PR. I'm inclined to push a change to `pandas_implementation.py` and then merge this.

rgommers:
> > With one change to make: rename `mapping` to `categories`,
> 
> This change was made in the spec part, but not the rest of this PR. I'm inclined to push a change to `pandas_implementation.py` and then merge this.

I tried this - since the tests no longer pass with this PR, and we now have an _actual_ Pandas implementation, it may be better to completely remove this early prototype in `pandas_implementation.py`.

rgommers:
> I tried this - since the tests no longer pass with this PR, and we now have an _actual_ Pandas implementation, it may be better to completely remove this early prototype in `pandas_implementation.py`.

Discussed in today's call: everyone is happy with deleting this prototype code.

rgommers (approved):
Everyone is happy and this is now a pretty small/straightforward diff. So merging. I'll follow up with all the implementers to ensure we will actually propagate these changes to all implementations.

Thanks @shwina, @vnlitvinov, @kkraus14

PR #70, opened 2021-10-29
Refactoring propositions

iskode:
I've made a few changes based on the feedback from the [cudf implementation PR](https://github.com/rapidsai/cudf/pull/9071#issuecomment-950140232) I've submit still under review.
There  3 main refactoring (one for each commit):

1. The first part of the `buffer_to_ndarray` method is convert the protocol dtype to numpy one. I've refactored it in a dedicated method as in cudf implementation, I needed it a dozen of times so this might be the case for pandas in near future. Also this method used some local constants that I've moved into module level constant. This will prevent to recreate these variables for each call.
2. I've moved the Device class from the`__dlpack_device__` method to the module level as `_DtypeKind`. Then I've add all the devices values so that the code will show which are currently support and maintainers will know directly about future device supports.
3. I've noticed that `convert_string_column` return all buffers while `buffer_to_ndarray` and `convert_categorical_column` return only the data buffer. So I've set these latter to return all buffers as well.

### DIFF
--- a/protocol/pandas_implementation.py
+++ b/protocol/pandas_implementation.py
@@ -122,9 +122,12 @@ def convert_column_to_ndarray(col : ColumnObject) -> np.ndarray:
     if col.describe_null[0] not in (0, 1):
         raise NotImplementedError("Null values represented as masks or "
                                   "sentinel values not handled yet")
-
-    _buffer, _dtype = col.get_buffers()["data"]
-    return buffer_to_ndarray(_buffer, _dtype), _buffer
+    buffers = col.get_buffers()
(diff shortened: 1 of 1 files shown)


rgommers:
Thanks a lot @iskode. Given that after gh-81 the implementation has been removed from this repo, and Pandas now has its own implementation merged (with significant improvements compared to the prototype that this PR was refactoring), I will close this PR. 

We did learn quite a bit from your cuDF implementation work. Thanks for opening this PR, and apologies it did not get merged last year.

iskode:
No problem. You don't have to apologies !
Thank you for your mentoring. It has been very great experience.

PR #72, opened 2021-12-28
Fix link to seaborn docs

mwaskom:


rgommers (approved):
Oops, simple copy-paste error there. Thanks for the fix @mwaskom

PR #81, opened 2022-07-14
Remove prototype Pandas implementation from `protocol/`

rgommers:
Now that we have four real-world implementations in cuDF, Vaex, Modin and Pandas, we no longer need this prototype. Having it here makes it harder to merge other PRs (see gh-69), so let's remove it now.


rgommers:
No futher comments, and folks were happy with this change in the last call. So I'll go ahead and merge this.

PR #87, opened 2022-09-24
Clarify dtype type in `ColumnBuffers`  elements

flexatone:
Thanks for this team's efforts in implementing the dataframe interchange protocol.

While experimenting on an implementation for StaticFrame (https://github.com/InvestmentSystems/static-frame/pull/621), I found one aspect of the ABC and associated types unclear.

The `ColumnBuffers` `TypedDict` defines its elements as pairs of `Buffer` and a dtype, but only types the dtype as `Any`. I assume that the dtype should be provided as the same tuple returned from `Column.dtype`. This PR makes this explicit by defining a `Dtype` type alias and using it for the return value of `Column.dtype` as well as the elements in `ColumnBuffers`.

If my assumption is incorrect, it would be helpful to specify the type of the dtype values returned from `ColumnBuffers`.

### DIFF
--- a/protocol/dataframe_protocol.py
+++ b/protocol/dataframe_protocol.py
@@ -58,6 +58,9 @@ class DtypeKind(enum.IntEnum):
     CATEGORICAL = 23
 
 
+Dtype = Tuple[DtypeKind, int, str, str]
+
+
 class ColumnNullType(enum.IntEnum):
     """
     Integer enum for null type representation.
@@ -91,7 +94,7 @@ class ColumnBuffers(TypedDict):
     # first element is a buffer containing mask values indicating missing data;
     # second element is the mask value buffer's associated dtype.
     # None if the null representation is not a bit or byte mask
-    validity: Optional[Tuple["Buffer", Any]]
+    validity: Optional[Tuple["Buffer", Dtype]]
 
     # first element is a buffer containing the offset values for
     # variable-size binary data (e.g., variable-length strings);
@@ -237,7 +240,7 @@ class Column(ABC):
 
     @property
     @abstractmethod
-    def dtype(self) -> Tuple[DtypeKind, int, str, str]:
+    def dtype(self) -> Dtype:
         """
         Dtype description as a tuple ``(kind, bit-width, format string, endianness)``.
 
--- a/protocol/dataframe_protocol.py
+++ b/protocol/dataframe_protocol.py
@@ -89,7 +89,7 @@ class ColumnNullType(enum.IntEnum):
 class ColumnBuffers(TypedDict):
     # first element is a buffer containing the column data;
     # second element is the data buffer's associated dtype
-    data: Tuple["Buffer", Any]
+    data: Tuple["Buffer", Dtype]
 
     # first element is a buffer containing mask values indicating missing data;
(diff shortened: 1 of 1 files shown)


protocol/dataframe_protocol.py
```diff
@@ -58,6 +58,9 @@ class DtypeKind(enum.IntEnum):
     CATEGORICAL = 23
 
 
+Dtype = Tuple[DtypeKind, int, str, str]
```
honno, line 61:
```suggestion
Dtype = Tuple[DtypeKind, int, str, str]  # see Column.dtype
```

Could also maybe move the `Column.dtype` description here as evidently the dtype tuple gets returned elsewhere, but eh probably nicer to keep things as-is.

honno (commented):
Thanks, I think this would be useful.

> If my assumption is incorrect

Just as a sanity check I checked and pandas/vaex/cuDF/modin all return the type described in `Column.dtype` for the `get_buffers()` values. <sup>A. TODO I realize for `dataframe-interchange-tests` is generalise [`test_dtype`](https://github.com/data-apis/dataframe-interchange-tests/blob/a9ff93fcf8a0cb4990801a8f15f7334d9a79fb6c/tests/test_column_object.py#L70) and use it in [`test_get_buffers`](https://github.com/data-apis/dataframe-interchange-tests/blob/a9ff93fcf8a0cb4990801a8f15f7334d9a79fb6c/tests/test_column_object.py#L188).</sup>

protocol/dataframe_protocol.py
flexatone: Yeah, I wanted to put the `Column.dtype` docs closer to the type alias definition but it is a little awkward. The comment that you suggest is good compromise.

rgommers (approved):
Thanks @flexatone, this looks correct to me and is a useful improvement. And thanks @honno for the review.

Also nice to learn about StaticFrame!

jorisvandenbossche:
> > If my assumption is incorrect
> 
> Just as a sanity check I checked and pandas/vaex/cuDF/modin all return the type described in `Column.dtype` for the `get_buffers()` values. A. TODO I realize for `dataframe-interchange-tests` is generalise [`test_dtype`](https://github.com/data-apis/dataframe-interchange-tests/blob/a9ff93fcf8a0cb4990801a8f15f7334d9a79fb6c/tests/test_column_object.py#L70) and use it in [`test_get_buffers`](https://github.com/data-apis/dataframe-interchange-tests/blob/a9ff93fcf8a0cb4990801a8f15f7334d9a79fb6c/tests/test_column_object.py#L188).

Note that this is actually _not_ correct, depending on how you interpret it. Yes, the buffers' dtype returns a similar _type_ of DType tuple, but it should not necessarily return the _same_ dtype tuple as its `Column.dtype` does, as the buffer can have a different dtype than the column. 

It seems that we all interpreted this wrongly and all implementations got this wrong (or the text about _"the data buffer's associated dtype"_ is wrong), see https://github.com/apache/arrow/issues/37598, https://github.com/pandas-dev/pandas/issues/54781, https://github.com/pola-rs/polars/pull/10787 (and the same for StaticFrame mentioned above, from a quick look).

PR #92, opened 2022-12-13
Update Sphinx conf.py for the dataframe API site to use Myst

rgommers:


jbrockmendel:
looks benign to me.  should we ask someone more familiar with sphinx to take a look or is this sufficiently easy to just merge?

rgommers:
Thanks for having a look @jbrockmendel. It's sufficiently easy I'd say. And it matches what we use for https://data-apis.org/array-api/latest/, so I'm pretty sure it works as advertised.

PR #94, opened 2022-12-20
Add indexing, assign, arithmetic methods

jbrockmendel:


spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -1,5 +1,338 @@
...
+from typing import Sequence, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from .column_object import Column
+
 
 class DataFrame:
-    pass
+
+    def get_column_by_name(self, key: str) -> Column:
```
rgommers, line 11: minor: `name` seems a little more logical than `key` to me. Also, that is what it's called in the interchange protocol. Now that I'm thinking about this though: how about making this positional-only? That results in more idiomatic code, and hence seems like a good idea to me independent of whether then argument name is `key` or `name`. So I propose:
```python
def get_column_by_name(self, name: str, /) -> Column:
```

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -1,5 +1,338 @@
...
+        Column
+
+        Raises
+        ------
+        KeyError
+            If the key is not present.
+        """
+        ...
+
+    def get_columns_by_name(self, keys: Sequence[str]) -> DataFrame:
```
rgommers, line 30: The type annotation should be `"DataFrame"` (string quotes)

spec/API_specification/dataframe_api/dataframe_object.py
rgommers: comment applies to `get_columns_by_name` as well.

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -1,5 +1,338 @@
...
+            Positions of rows to select.
+
+        Returns
+        -------
+        DataFrame
+
+        Notes
+        -----
+        Some discussion participants prefer a stricter type Column[int] for
+        indices.
```
rgommers, line 65: I suggest to add "in order to make it easier to implement with good performance on GPUs".

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -1,5 +1,338 @@
...
+        DataFrame
+
+        Notes
+        -----
+        Some participants preferred a weaker type Arraylike[bool] for mask,
+        where 'Arraylike' denotes an object adhering to the Array API standard.
+        """
+        ...
+
+    def insert(self, loc: int, label: str, value: Column) -> DataFrame:
```
rgommers, line 106: Missing description. The one thing that isn't obvious to me is what values `loc` can take. The pandas docs say _"Insertion index. Must verify `0 <= loc <= len(columns)`."_ So no negative values with wrap-around allowed (seems fine to me).

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -1,5 +1,338 @@
...
+        """
+        ...
+
+    def __eq__(self, other: DataFrame | "Scalar") -> DataFrame:
+        """
+        Parameters
+        ----------
+        other : DataFrame or Scalar
+            If DataFrame, must have same length and matching columns.
+            "Scalar" here is defined implicitly by the contained dtypes.
```
rgommers, line 149: The one thing to clarify here is dtype casting rules. For numerical dtypes, follow the regular casting rules (matching array API standard), and for other dtypes no upcasting, dtypes must match exactly?

rgommers (commented):
Thanks @jbrockmendel, this looks pretty good to me. A few questions and comments inline. The one important one is about casting rules.

To other reviewers: the idea is to merge this PR (and the following ones) relatively quickly. This not to finalize decisions, but to construct a complete API that is ready for broader review as an RFC.

spec/API_specification/dataframe_api/dataframe_object.py
jbrockmendel: sure.

Whats the reasoning for the trailing "/"?

rgommers: To have one way of spelling things instead of two. Right now:
- `df.get_column_by_name('a_name')`
- `df.get_column_by_name(key='a_name')` (or `df.get_column_by_name(name='a_name')` 

It looks to me like the second form is redundant and less readable, so why support it? Both positional-only and keyword-only arguments are great in many cases where the argument name doesn't add anything.

jbrockmendel: ok, updated.

spec/API_specification/dataframe_api/dataframe_object.py
rgommers: I guess we can leave this as is for now. Documenting how casting rules work should be a separate PR, and then we can refer to that one description from all these docstrings.

rgommers (approved):
LGTM, in it goes. Thanks @jbrockmendel

PR #95, opened 2022-12-21
API: groupby reductions

jbrockmendel:


rgommers (commented):
Overall this looks good to me I think. For common understanding, let me try to summarize the design decisions that I think went into this fairly minimal API:
- The `groupby` method should only accept a list of keys. No single key or other selectors.
   - functions/expressions in groupby would be equivalent to adding a separate column with the result of the expression, and grouping on that.
- Returns a new `Groupby` oject, with all methods being reductions that will match the reduction methods on `DataFrame`.

`skipna` would be good to explain. Would that have the same semantics as `dropna` in pandas?

jbrockmendel:
> Would that have the same semantics as dropna in pandas?

Same semantics as "skipna" in pandas groupby/ndframe reductions

rgommers (approved):
Fixed merge conflict, LGTM now. In it goes. Thanks @jbrockmendel

PR #96, opened 2022-12-23
API: isnull, isnan, reductions

jbrockmendel:


jbrockmendel:
@rgommers i think after this theres a) dropna/fillna (which are tough bc the notes got lost from the hackmd doc) and b) a briefly-discussed `where`.  Am I forgetting anything?

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -4,9 +4,15 @@
 
 if TYPE_CHECKING:
     from .column_object import Column
+    from .groupby_object import GroupBy
 
 
 class DataFrame:
+    def groupby(self, keys: list[str], /) -> GroupBy:
```
rgommers, line 11: This matches what was done/discussed before, see gh-95

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -4,9 +4,15 @@
 
 if TYPE_CHECKING:
     from .column_object import Column
+    from .groupby_object import GroupBy
```
rgommers, line 7: To address in a follow-up: this results in a circular import, so it needs a bit of refactoring it looks like.

rgommers (approved):
Thanks @jbrockmendel, and thanks for the patience waiting for a review. This LGTM overall. The reductions match those on the `GroupBy` object (or, more accurately, those on `GroupBy` match the ones on `DataFrame`). It may be useful to add comments to the `GroupBy` methods that for any details, the matching `DataFrame` methods should be used.

The `isnan` and `isnull` signatures also seem fine to me. The previous discussion left a question mark on the method name for `isnull` (e.g., could have been `ismissing`). But that's a bikeshed that can be painted in a future review.

I think I'll merge this as is. Perhaps you can open a follow up PR with the two tweaks I suggested (cross-reference reduction methods; circular import fix)?

PR #98, opened 2023-01-17
Render API docstrings in `.py` files as html

rgommers:
This does the following:
- Ensure the package in `spec/API_specification/dataframe_api/` is importable
- Ensure all docstrings for objects/methods are rendered in the html docs via `autodoc`
- Add a description for operators supported by the dataframe object (and a few TODOs)
- Adds Sphinx templates for styling the html rendered versions of methods and attributes properly (taken over from the array-api repo)
- Adds some infrastructure for static typing that we may need. So far only the `Scalar` class needed a `TypeVar`, but there should be more to do there.

### DIFF
--- a/spec/API_specification/column_object.rst
+++ b/spec/API_specification/column_object.rst
@@ -1,0 +1,23 @@
+.. _column-object:
+
+Column object
+=============
+
+A conforming implementation of the dataframe API standard must provide and
+support a column object having the following attributes and methods.
+
+-------------------------------------------------
+
+Methods
+-------
+TODO
+
+..
+  NOTE: please keep the methods in alphabetical order
+
+    .. currentmodule:: dataframe_api
+
+    .. autosummary::
+       :toctree: generated
+   :template: property.rst
+
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -4,12 +4,10 @@ from typing import Sequence, Union, TYPE_CHECKING
 from .column_object import Column
 from .groupby_object import GroupBy
 
+from ._types import Scalar
+
 
 __all__ = ["DataFrame"]
-
-
-class Scalar:
-    "A class to represent Python scalars"
 
 
 class DataFrame:
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -165,6 +165,8 @@ class DataFrame:
 
     def __eq__(self, other: DataFrame | "Scalar") -> DataFrame:
         """
+        Compare for equality.
+
         Parameters
         ----------
         other : DataFrame or Scalar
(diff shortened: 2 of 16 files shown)


spec/API_specification/column_object.rst
```diff
@@ -0,0 +1,23 @@
+.. _column-object:
+
+Column object
+=============
+
+A conforming implementation of the dataframe API standard must provide and
+support a column object having the following attributes and methods.
```
jbrockmendel, line 7: should column be capitalized?

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -1,10 +1,13 @@
-__all__ = ["DataFrame"]
+from __future__ import annotations
+from typing import Sequence, Union, TYPE_CHECKING
+
+from .column_object import Column
+from .groupby_object import GroupBy
```
jbrockmendel, line 5: why doing this instead of hiding it being TYPE_CHECKING?

jbrockmendel: (which is now imported but not used)

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -158,8 +161,10 @@ def set_column(self, label: str, value: Column) -> "DataFrame":
         """
         ...
 
-    def __eq__(self, other: DataFrame | "Scalar") -> "DataFrame":
+    def __eq__(self, other: DataFrame | "Scalar") -> DataFrame:
```
jbrockmendel, line 164: should "Scalar" be de-quoted?

jbrockmendel:
a few nitpicks, otherwise looks fine

spec/API_specification/dataframe_api/dataframe_object.py
rgommers: `import dataframe_api` was actually choking on the annotation, strangely enough. I'll reconfirm that by moving it back into `TYPE_CHECKING`, to make sure I wasn't mixing it up with another issue.

spec/API_specification/dataframe_api/dataframe_object.py
rgommers: The `|` syntax unfortunately doesn't work without the quotes.

spec/API_specification/dataframe_api/dataframe_object.py
rgommers: Okay, it was something else - I reverted this. `autodoc` debugging .... sigh

spec/API_specification/dataframe_api/dataframe_object.py
rgommers: Turns out that does work too. Adopted your suggestion.

spec/API_specification/column_object.rst
rgommers: I didn't do that on purpose. Maybe one thing to discuss in the context of constructors is whether the name matters (`Column`, `DataFrame`, etc.). For array libraries we left this out on purpose, because it's never needed and existing libraries name these things differently. The same could apply here, unless you allow calling `object.__init__` as a constructor.

rgommers:
CI for deploying is happy, rendered docs are up on https://data-apis.org/dataframe-api/draft/API_specification/index.html.

PR #101, opened 2023-02-13
Add polars to ecosystem

MarcoGorelli:


rgommers (commented):
Thanks @MarcoGorelli! Polars should definitely be added to this list, it's an important library. I think this content was written around the time of the very first commit in the Polars repo, which explains why it was missing.

In terms of the content, maybe it should be reworded a bit more neutrally? Other libraries like Vaex and Arrow are also very fast, but we didn't take the literal wording from their websites.

I'd also mention that it has both a Rust and a Python API.

rgommers (approved):
Thanks Marco. This could have been merged a while ago, sorry; let's get it in now.

PR #104, opened 2023-03-01
Update groupby docstring to make orderedness explicitly not guaranteed

kkraus14:
Updates the `DataFrame.groupby` docstring to clarify that orderdedness of both keys and rows within each group is not guaranteed and is implementation defined.

Additionally, changes keys to be a `Sequence[str]` to be in line with other APIs expecting a sequence of column names.

### DIFF
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -11,9 +11,27 @@ __all__ = ["DataFrame"]
 
 
 class DataFrame:
-    def groupby(self, keys: list[str], /) -> GroupBy:
+    def groupby(self, keys: Sequence[str], /) -> GroupBy:
         """
         Group the DataFrame by the given columns.
+
+        Parameters
+        ----------
+        keys : Sequence[str]
+
+        Returns
+        -------
+        GroupBy
+
+        Raises
+        ------
+        KeyError
+            If any of the requested keys are not present.
+
+        Notes
+        -----
+        The order of the keys and the order of rows within each group is not
+        guaranteed and is implementation defined.
         """
         ...
 
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -30,8 +30,9 @@ class DataFrame:
 
         Notes
         -----
-        The order of the keys and the order of rows within each group is not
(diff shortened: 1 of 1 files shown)


spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -11,9 +11,27 @@
...
+        GroupBy
+
+        Raises
+        ------
+        KeyError
+            If any of the requested keys are not present.
+
+        Notes
+        -----
+        The order of the keys and the order of rows within each group is not
```
jbrockmendel, line 33: I think this likely needs to be on the GroupBy class-level docstring, since this is referring to the output of GroupBy methods

kkraus14: Do we expect people to construct `GroupBy` objects directly? My thought was that this documentation would be most impactful where developers would typically look.

jbrockmendel: fair enough w/r/t location.  wording-wise id like to clarify that it is about results of aggregations, otherwise we're leaking implementation details about keys/groups

kkraus14: Let me take a stab at rewording it.

kkraus14: Tried rewording but still not very happy with it.

rgommers: It reads okay to me, but maybe it's useful to add something like: _In particular, sorting on `keys` is not required._? Just to insert "sort" somewhere

rgommers: That doesn't sound right. And it won't be just cuDF. For the record, this was basically the top wish list item (or pain point) from the SIG folks as well early on - they used hash-based methods too in their dataframe library.

jbrockmendel: I'm not suggesting wording that would suggest this is specific to pandas/cudf, just that an example may clarify what behavior should not be relied on.

rgommers: In the meeting there was a "ship it" sentiment - so let's do just that here:)

rgommers:
Thanks @kkraus14 and @jbrockmendel!

PR #106, opened 2023-03-14
Add rename columns and get column names

### LINKED ISSUE #105: DataFrame.rename_columns, DataFrame.column_names
Any reason to not have a `DataFrame.rename_columns` method, to allow for changing columns' names?

It would be pretty important if the Standard were to be used in [pyjanitor's clean names](https://github.com/pyjanitor-devs/pyjanitor/blob/e307216ce86fb223718c77d810c7872f52f31495/janitor/functions/clean_names.py#L84-L107)

The spec could be:
```python
    def rename_columns(self, mapping: Mapping[str, str]) -> DataFrame:
        """
        Rename columns.

        Parameters
        ----------
        mapping
            Mapping where keys are old column names and values are new column names.

        Returns
        -------
        DataFrame
        """
        ...
```

The other useful one would be `DataFrame.column_names`.

Then, pyjanitor's `clean_names` could be rewritten as something like

```python
def clean_names_with_standard(
    df,
    strip_underscores=None,
    case_type='lower',
    remove_special=False,
    strip_accents=True,
    preserve_original_columns=True,
    enforce_string=True,
    truncate_limit=None,
):
    df = StandardDataFrame(df)

    mapping = {}
    for old_label in df.column_names:
        new_label = _change_case(old_label, case_type)
        new_label = _normalize_1(new_label)
        if remove_special:
            new_label = _remove_special(new_label)
        if strip_accents:
            new_label = _strip_accents(new_label)
        new_label = re.sub("_+", "_", new_label)
        new_label = _strip_underscores_func(new_label, strip_underscores)
        new_label = new_label[:truncate_limit]
        mapping[old_label] = new_label
    df = df.rename(mapping)

    # return the underlying DataFrame
    # the user needn't care that the Standard was used under-the-hood
    return df.df
```

and it should work on any DataFrame which has an implementation of the standard

Example:
```python
>>> pd_df
   Aloha  Bell Chart  Animals@#$%^
0      0           0             0
1      1           1             1
2      2           2             2

>>> clean_names_with_standard(pd_df)
   aloha  bell_chart  animals@#$%^
0      0           0             0
1      1           1             1
2      2           2             2

>>> pl_df
shape: (3, 3)
┌───────┬────────────┬──────────────┐
│ Aloha ┆ Bell Chart ┆ Animals@#$%^ │
│ ---   ┆ ---        ┆ ---          │
│ i64   ┆ i64        ┆ i64          │
╞═══════╪════════════╪══════════════╡
│ 0     ┆ 0          ┆ 0            │
│ 1     ┆ 1          ┆ 1            │
│ 2     ┆ 2          ┆ 2            │
└───────┴────────────┴──────────────┘

>>> clean_names_with_standard(pl_df)
shape: (3, 3)
┌───────┬────────────┬──────────────┐
│ aloha ┆ bell_chart ┆ animals@#$%^ │
│ ---   ┆ ---        ┆ ---          │
│ i64   ┆ i64        ┆ i64          │
╞═══════╪════════════╪══════════════╡
│ 0     ┆ 0          ┆ 0            │
│ 1     ┆ 1          ┆ 1            │
│ 2     ┆ 2          ┆ 2            │
└───────┴────────────┴──────────────┘
```

MarcoGorelli:
closes #105 

Doesn't look like there's objections, so I went ahead and opened this

### DIFF
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -195,7 +195,7 @@ class DataFrame:
         """
         ...
 
-    def get_column_names(self) -> tuple[str, ...]:
+    def get_column_names(self) -> Sequence[str]:
         """
         Get column names.
 
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -180,7 +180,7 @@ class DataFrame:
         """
         ...
 
-    def rename_columns(self, mapping: dict[str, str]) -> DataFrame:
+    def rename_columns(self, mapping: Mapping[str, str]) -> DataFrame:
         """
         Rename columns.
 
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -201,7 +201,7 @@ class DataFrame:
 
         Returns
         -------
-        tuple[str, ...]
+        Sequence[str]
         """
         ...
 
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -186,7 +186,7 @@ class DataFrame:
 
         Parameters
         ----------
-        mapping : dict[str, str]
+        mapping : Mapping[str, str]
             Keys are old column names, values are new column names.
 
         Returns
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -195,13 +195,13 @@ class DataFrame:
         """
         ...
 
-    def get_column_names(self) -> list[str]:
+    def get_column_names(self) -> tuple[str, ...]:
         """
         Get column names.
 
         Returns
         -------
-        list[str]
+        tuple[str, ...]
         """
         ...
 
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -180,6 +180,31 @@ class DataFrame:
         """
         ...
 
+    def rename_columns(self, mapping: dict[str, str]) -> DataFrame:
+        """
+        Rename columns.
+
+        Parameters
+        ----------
+        mapping : dict[str, str]
+            Keys are old column names, values are new column names.
+
+        Returns
+        -------
+        DataFrame
+        """
+        ...
+
+    def get_column_names(self) -> list[str]:
+        """
+        Get column names.
+
+        Returns
+        -------
+        list[str]
+        """
+        ...
+
     def __eq__(self, other: DataFrame | Scalar) -> DataFrame:
         """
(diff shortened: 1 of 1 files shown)


rgommers (commented):
This LGTM, thanks @MarcoGorelli. One question: do we expect/allow users to mutate the returned list? If not, a tuple of strings may be nicer (can return an existing tuple perhaps, a list must be copied) and may be slightly nicer for a JIT compiler.

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -180,6 +180,31 @@ def set_column(self, label: str, value: Column) -> DataFrame:
...
+        """
+        ...
+
+    def get_column_names(self) -> list[str]:
+        """
+        Get column names.
+
+        Returns
+        -------
+        list[str]
```
MarcoGorelli, line 204:
```suggestion
    def get_column_names(self) -> tuple[str, ...]:
        """
        Get column names.

        Returns
        -------
        tuple[str, ...]
```

MarcoGorelli:
yup, sounds good, have updated

rgommers:
A bit more discussion on the return type:
- For really wide dataframes, @kkraus14 pointed out that a Python objects like a tuple may be limiting. We decided to punt on this though, since there's no clear alternative. Implementers could return some other object that has tuple-compatible semantics.
- Some discussion about allowing returning `Sequence[str, ...]` instead. Assuming Mypy is happy with that, that seems like an improvement.

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -180,6 +180,31 @@ def set_column(self, label: str, value: Column) -> DataFrame:
...
+        mapping : dict[str, str]
+            Keys are old column names, values are new column names.
+
+        Returns
+        -------
+        DataFrame
+        """
+        ...
+
+    def get_column_names(self) -> tuple[str, ...]:
```
MarcoGorelli, line 198:
```suggestion
    def get_column_names(self) -> Sequence[str]:
```

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -180,6 +180,31 @@ def set_column(self, label: str, value: Column) -> DataFrame:
...
+        """
+        Get column names.
+
+        Returns
+        -------
+        tuple[str, ...]
```
MarcoGorelli, line 204:
```suggestion
        Sequence[str]
```

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -180,6 +180,31 @@ def set_column(self, label: str, value: Column) -> DataFrame:
         """
         ...
 
+    def rename_columns(self, mapping: dict[str, str]) -> DataFrame:
+        """
+        Rename columns.
+
+        Parameters
+        ----------
+        mapping : dict[str, str]
```
kkraus14, line 189:
```suggestion
        mapping : Mapping[str, str]
```

kkraus14 (approved):

PR #110, opened 2023-03-20
add dataframe property

### LINKED ISSUE #108: Method to get underlying object
Does there need to be a way to get back the underlying object?

I'm thinking about the `pyjanitor` `clean_names` example

Some user starts with a DataFrame (say, a pandas one) `df`, and calls `clean_names(df)`. They would probably expect to get back what they started with, without caring that PyJanitor internally used the standard.

For example, PyJanitor could do
```
def clean_names(df, ...):
    df = dataframe_standard(df)  # or whatever the way to enable the standard will be
    df = ...  # clean names
    return df.dataframe  # return same type of DataFrame as was passed
```

So, should some `.dataframe` property be added, so that the library can "opt-out" of the standard once it has done all its work?

MarcoGorelli:
closes #108

### DIFF
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -11,6 +11,15 @@ __all__ = ["DataFrame"]
 
 
 class DataFrame:
+    @property
+    def dataframe(self) -> object:
+        """
+        Return underlying (not-necessarily-Standard-compliant) DataFrame.
+
+        If a library only implements the Standard, then this can return `self`.
+        """
+        ...
+
     def groupby(self, keys: Sequence[str], /) -> GroupBy:
         """
         Group the DataFrame by the given columns.


rgommers (commented):
Thanks Marco. One question: isn't it better to type "returns unknown object" with `-> Any` rather than `-> object`?

MarcoGorelli:
`-> object` would ensure that no properties of the object are being used - e.g.
```python
from typing import *

class Animal:
    def foo(self) -> object:
        ...
        
class Cat(Animal):
    def foo(self) -> int:
        return 3

reveal_type(Cat().foo())
```
type checks fine, and would error on `Animal().foo().sort_values()`, whereas a return type of `Any` would turn off the type checker and `Animal().foo().sort_values()` would pass (though that's probably not what we want, because there would be no guarantees about what methods are in the return value from `.dataframe`)

rgommers:
Ah fair enough. I haven't seen `-> object` used before, but the rationale seems solid here. And for users who know what type they want/expect, they can type-narrow themselves.

rgommers (approved):
This seems straightforward enough and the one minor concern on typing was answered - in it goes. Thanks @MarcoGorelli

PR #111, opened 2023-03-20
rename isnull and isnan methods

MarcoGorelli:
as discussed on the last call

### DIFF
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -478,7 +478,7 @@ class DataFrame:
         """
         ...
 
-    def isnull(self) -> DataFrame:
+    def is_null(self) -> DataFrame:
         """
         Check for 'missing' or 'null' entries.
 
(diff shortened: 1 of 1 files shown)


rgommers:
I don't remember what the rationale for this was, can you add one @MarcoGorelli? These methods are using the no-underscore flavor pretty much universally I think, and also in the array API for `isnan`.

MarcoGorelli:
just snake-case - you'd read it as "is null", and replacing the space with an underscore, would get "is_null"

rgommers:
Hmm - I'd think that that's probably not a good reason for `is` and `has`, they're kinda special in Python (`isinstance`, `issubclass`, `hasattr`, etc.) - that plus `isnan`/`isnull` being the current names everywhere means we should probably leave this as is.

jorisvandenbossche:
> everywhere

Just to be pedantic, but not "everywhere": for example pyarrow and polars both use "is_null" etc (but yeah, those are certainly the outliers and probably less relevant for the exercise here)

PR #112, opened 2023-03-20
Add from_dict classmethod

MarcoGorelli:
as discussed on the latest call

### DIFF
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -1,5 +1,5 @@
 from __future__ import annotations
-from typing import Sequence, Union, TYPE_CHECKING, NoReturn, Self, Mapping
+from typing import Sequence, Union, TYPE_CHECKING, NoReturn, Mapping
 
 if TYPE_CHECKING:
     from .column_object import Column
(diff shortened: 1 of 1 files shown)


MarcoGorelli:
Thinking about this again, I'm not sure how this would work

In order for this to be possible, I think the following would probably be necessary:
- constructors for `Column`
- a way to generate a given DataFrame from a dict of Columns, which would probably require being able to get the underlying data from the columns

Let's talk about this more next time, but I'm not sure `from_dict` is the solution

rgommers:
> constructors for `Column`

There should be a lot of functions that one can apply directly to columns, so you can get new columns from ops on existing ones. But agreed - probably also need to be able to construct columns from sequences and arrays for many use cases.

> a way to generate a given DataFrame from a dict of Columns, which would probably require being able to get the underlying data from the columns

This one I don't understand - this should be a completely trivial operation, and no access of underlying data should be needed (a dataframe is simply an ordered collection of columns). Or am I missing your point?

MarcoGorelli:
Since we agreed in https://github.com/data-apis/dataframe-api/issues/118 that each library can have private methods to access the underlying data of its own columns, you're right that it is indeed trivial - e.g. in pandas we could do
```python
    @classmethod
    def from_dict(cls, data: dict[str, PandasColumn]) -> PandasDataFrame:
        return cls(pd.DataFrame({label: column._series for label, column in data.items()}))
```

MarcoGorelli:
is this one OK, or does anything else need doing?

rgommers (approved):
This LGTM indeed, in it goes. Thanks Marco.

PR #113, opened 2023-03-20
poison __iter__

### LINKED ISSUE #103: Iterating over a column
In the last meeting, there was a section when we talked some default which, if pandas changed, would make things easier for other libraries.

One example which @kkraus14 gave, at about 0:43:30, was about iterating over a column in Python-land, value by value.

@shwina mentioned that polars doesn't allow this.

I may have misunderstood what Keith was referring to, but to me it looks like it is allowed?
```python
In [77]: df = pl.DataFrame({'ts': pl.date_range(datetime(2020, 1, 1), datetime(2020, 1, 10))})

In [78]: for i in range(len(df)):
    ...:     print(f'value at position {i} is: {df["ts"][i]}')
    ...:
value at position 0 is: 2020-01-01 00:00:00
value at position 1 is: 2020-01-02 00:00:00
value at position 2 is: 2020-01-03 00:00:00
value at position 3 is: 2020-01-04 00:00:00
value at position 4 is: 2020-01-05 00:00:00
value at position 5 is: 2020-01-06 00:00:00
value at position 6 is: 2020-01-07 00:00:00
value at position 7 is: 2020-01-08 00:00:00
value at position 8 is: 2020-01-09 00:00:00
value at position 9 is: 2020-01-10 00:00:00
```

Could you please clarify what it is that polars doesn't allow, which you would like it if pandas disallowed?

MarcoGorelli:
closes #103

### DIFF
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -1,5 +1,5 @@
 from __future__ import annotations
-from typing import Sequence, Union, TYPE_CHECKING
+from typing import Sequence, Union, TYPE_CHECKING, NoReturn
 
 if TYPE_CHECKING:
     from .column_object import Column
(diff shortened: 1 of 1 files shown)


rgommers (approved):
LGTM, thanks @MarcoGorelli!

kkraus14 (approved):

rgommers:
No further objections, so in it goes

PR #119, opened 2023-03-23
add DataFrame.pivot

MarcoGorelli:
Let's discuss this one in the next call.

One open question I have is how to name the resulting columns. I'm suggesting to do something like .`{value_0}_{label_0}_{label_1}..._{label_n}`

For a simple example: say we start with
```python
a b c d
1 1 x p
4 2 x q
```
and call `df.pivot(values=['a'], keys=['b'], columns=['c', 'd'])`. We would get:
```python
b a_x_p a_x_q
1     1  null     
2  null     4
```

### DIFF
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -539,3 +539,47 @@ class DataFrame:
         Does *not* include 'missing' or 'null' entries.
         """
         ...
+
+    def pivot(self, values: Sequence[str], keys: Sequence[str], labels: Sequence[str]) -> DataFrame:
+        """
+        Reshape data into columns (from rows).
+
+        Does not support aggregation: each row must have a unique value for
+        (`keys`, `labels`). If that's not the case, then the caller is
+        responsible for de-duplicating based on (`keys`, `labels`) beforehand.
+
+        The result will have one column for each element of `keys`, as well as one
+        column for each unique combination of each element of `values`, each element
+        of `labels`, and each value in each element of `labels`.
+        As no MultiIndex is available, multiple levels will have their names
+        separated by underscores.
+
+        There are four cases to consider:
+        - single `values`, single `labels`:
+          columns will be of the form `{label_value}`;
+        - multiple `values`, single `labels`:
+          columns will be of the form `{value_name}_{label_value}`;
+        - single `values`, multiple `labels`:
+          columns will be of the format `{label_name}_{label_value}`;
+        - multiple `values`, multiple `labels`:
+          columns will be of the format `{value_name}_{label_name}_{label_value}`.
+
+        Parameters
+        ----------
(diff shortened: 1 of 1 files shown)


rgommers:
Thanks Marco! One question I have is whether this `pivot` API guarantees a predictable output shape. The way I see it in pandas and polars, the number of rows and columns in the result depends on the _unique values_ of `columns` and `values` respectively. You don't say whether `labels` is supposed to be unique. If it is, then that should be handled in user code explicitly it seems. And if not, what happens when there are clashes?

MarcoGorelli:
> The way I see it in pandas and polars, the number of rows and columns in the result depends on the unique values of columns and values respectively.

Not quite - say we start with a DataFrame `df`, and call `df.pivot(values, keys, columns)`:

- number of rows will equal `len(df.drop_duplicates(index))`
- the number of columns should equal to `len(keys) + len(values) * len(df.drop_duplicates(columns)`

The unique values of `values` shouldn't matter

> You don't say whether labels is supposed to be unique

Passing `labels=['a', 'a', 'b']` would be disallowed - thank, I'll clarify that.
But passing `labels=['a', 'b']`, where `'a'` is a column which has duplicate values, would be allowed. All that's required is that each row has a unique combination of `(keys, labels)`.
For example, if we call `df.pivot(values=['a', 'b'], keys=['c', 'd'], columns=['e', 'f'])`, then each row should have a unique combination of `('c', 'd', 'e', 'f')`. This is because this row's value for `'a'` would populate the cell in the result corresponding to the row given by its values for `('c', 'd')` and column given by `('e', 'f')`

rgommers:
Thanks, that helps!

> If there were multiple rows corresponding to that `('c', 'd', 'e', 'f')` combination, then they would need aggregating - but I'm suggesting we explicitly rule that out here

So in case a user of this API can't be sure whether or not there may be duplicates, they'd always call `.agg(...).pivot(...)` to avoid getting an exception?

MarcoGorelli:
Yes, exactly - I'm suggesting that they explicitly aggregate their data before hand (which should be possible with the groupby methods in the standard)

MarcoGorelli (commented):
Going to block myself on this one for now

As discussed yesterday, it might be a good idea to leave this one out of v1 until we have a stronger need for it

Trying out seaborn, their most standard plots such as
```python
import seaborn as sns
flights = sns.load_dataset("flights")

may_flights = flights.query("month == 'May'")
sns.lineplot(data=may_flights, x="year", y="passengers")
```
don't go through the `pivot` codepath

It's unlikely we'd be able to support seaborn in one go anyway, so this means that the plots which do use pivot would be unavailable for now, I think that's OK

MarcoGorelli:
one issue with `pivot` (in general) is that, the schema being value-dependent, it doesn't really lend itself to lazy computation

I suggest we just leave it out for now then - closing then

PR #127, opened 2023-04-03
add any and all rowwise

### LINKED ISSUE #124: boolean row-wise reductions?
Currently these can be done, but they're a bit clunky

e.g. of how one might do `dropna` using the current api:
```python
columns = df.get_column_names()
mask = ~df.get_column_by_name(columns[0]).isnan()
for col in columns[1:]:
    mask &= ~df.get_column_by_name(col).isnan()
result = df.get_rows_by_mask(mask)
```

It might be good to have something that takes a boolean DataFrame, like
```
df.isnan()
```
and returns a 1-D Column which reduced the DataFrame by row, either with `any` or `all`

OK with not having `DataFrame.mean(axis=1)` as part of the standard, but for boolean values this seems like a fairly important task for filtering?

e.g. of where I came across this:

https://github.com/mwaskom/seaborn/blob/5d9f37159bbd3ac44c8c8a06825583ba25648525/seaborn/relational.py#L520

MarcoGorelli:
closes #124

### DIFF
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -475,6 +475,34 @@ class DataFrame:
         Reduction returns a 1-row DataFrame.
         """
         ...
+    
+    def any_rowwise(self, skipna: bool = True) -> Column:
+        """
+        Reduction returns a Column.
+
+        Differs from ``DataFrame.any`` and that the reduction happens
+        for each row, rather than for each column.
+
+        Raises
+        ------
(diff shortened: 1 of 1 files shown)


rgommers (approved):
LGTM, thanks @MarcoGorelli.

I'll leave it open for a bit in case anyone else wants to have a look. The one possible bikeshed is the exception type - if a column dtype is wrong, it could be `ValueError` or `TypeError`. Both are arguably reasonable, I don't have a clear preference.

PR #129, opened 2023-04-03
add groupby.size

### LINKED ISSUE #122: GroupBy.size
Using GroupBy.size was suggested as an alternative to having `value_counts`, but it still needs adding

MarcoGorelli:
closes #122


rgommers (approved):
LGTM, and agrees with the discussion in gh-122 (in particular, leaving out `skip_na`). Thanks Marco!

PR #130, opened 2023-04-04
add shape property

MarcoGorelli:


spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -19,6 +19,12 @@ def dataframe(self) -> object:
         If a library only implements the Standard, then this can return `self`.
         """
         ...
+    
+    @property
+    def shape(self) -> tuple[int, int]:
+        """
+        Return number of rows and number of columns.
+        """
```
kkraus14, line 27: I think in the past we discussed that to better support lazy implementations that we'd prefer to have length and shape be functions instead of properties to better indicate that it may not be a free calculation

MarcoGorelli: sure, have updated

kkraus14 (approved):

PR #138, opened 2023-04-17
remove set_column

### LINKED ISSUE #116: Why does `set_column` exist?
I find this method a bit odd for two reasons.

First, it does two things: either replaces a column, or inserts a new one. This seems at odds with the rest of the Standard, which is typically quite strict and minimal.

Second, it can be implemented in terms of other Standard methods in just a few lines of code:
```python
    def set_column(self, label: str, value: Column) -> DataFrame:
        columns = self.get_column_names()
        if label in columns:
            idx = columns.index(label)
            return self.drop_column(label).insert(idx, label, value)
        return self.insert(len(columns), label, value)
```
Is this right?

If so, I'd suggest dropping it. Otherwise, if we're going to keep it just because it's convenient, then I'd argue that we should also have `value_counts`

MarcoGorelli:
As discussed in the last call, let's keep this one out for now

closes #116

### DIFF
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -174,21 +174,6 @@ class DataFrame:
         """
         ...
 
-    def set_column(self, label: str, value: Column) -> DataFrame:
-        """
-        Add or replace a column.
-
-        Parameters
-        ----------
-        label : str
-        value : Column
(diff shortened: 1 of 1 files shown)


kkraus14 (approved):

rgommers (approved):
Okay, let's get this one in. Thanks Marco

Issue #89, opened 2022-10-18
API: Indexing


jbrockmendel:
Summarizing the [discussion](https://hackmd.io/EiGt1U6RQbWFR6nIznDSgw#API-design-for-indexing) from the call 2022-10-13, the agreed-upon API looks like:

```
def get_column_by_name(self, key: str) -> Column: ...
def get_columns_by_name(self, keys: Sequence[str]) -> DataFrame: ...
def get_rows(self, indices: Punting[int]) -> DataFrame: ...
def slice_rows(self, start: int | None, stop: int | None, step: int | None) -> DataFrame: ...
def get_rows_by_mask(self, mask: Punting[bool]) -> DataFrame: ...
```

`Punting[x]` indicates that we discussed several options and decided to put off making a decision.  For `get_rows` the options were `Sequence[int]` and `Column[int]`.  For `get_rows_by_mask` the options were `Arraylike[bool]` and `Column[bool]` (here Arraylike _probably_ denotes an object adhering to the array API standard, but some participants wanted to be stricter than that).

The following indexing methods from pandas.DataFrame are _not_ included:

```
__getitem__
take
loc
iloc
at
iat
xs
```

rgommers:
The above-listed APIs for indexing were implemented: https://github.com/data-apis/dataframe-api/blob/705908137c6421a602765f53a3fbc6ae301901a9/spec/API_specification/dataframe_api/dataframe_object.py#L96-L183

So I'll close this issue.

PR #140, opened 2023-04-17
add __len__ and __getitem__ to Column

MarcoGorelli:


rgommers (approved):
LGTM, thanks @MarcoGorelli. Let's leave it open for a couple of days to give others a chance to comment.

spec/API_specification/dataframe_api/column_object.py
```diff
@@ -1,2 +1,24 @@
...
-    pass
+    def __len__(self) -> int:
+        """
+        Return the number of rows.
+        """
+    
+    def __getitem__(self, row) -> object:
+        """
+        Get the element at row index `row`.
+        """
```
kkraus14, line 12: What is the expected type of "row" here? An integer?

I wonder if we should have more explicit methods than `__getitem__` since intuitively people will try to feed both scalars and Sequences / Columns into it.

MarcoGorelli: oops, forgot to annotate that one, thanks

MarcoGorelli: > I wonder if we should have more explicit methods than __getitem__ since intuitively people will try to feed both scalars and Sequences / Columns into it.

could you expand please?

kkraus14: Historically, Pandas and other dataframe libraries have allowed passing scalar-like objects here and getting a scalar-like object in return or passing Sequence-like / Column-like objects here and getting Column-like objects in return. I.E. for Pandas:
```python
import pandas as pd

pds = pd.Series(['a', 'b', 'c', 'd', 'e'])

print(pds[0])  # Scalar-like input --> Scalar-like output
print(pds[[0]])  # Sequence-like input --> Column-like output
```

For our DataFrame object we have the `get_rows` method that takes a Sequence-like (should probably take a Column-like) where we could presumably have the same function for our Column object, and then instead of using `__getitem__` we could have something like `get_element` or `get_scalar` or something like that.

If we wanted to keep the `__getitem__` method I would suggest we handle both the Scalar-like and Sequence-like / Column-like cases as opposed to just the Scalar-like.

MarcoGorelli: > we have the get_rows method that takes a Sequence-like (should probably take a Column-like)

yeah, I came across this - I'll open a separate issue

MarcoGorelli: > If we wanted to keep the __getitem__ method I would suggest we handle both the Scalar-like and Sequence-like / Column-like cases as opposed to just the Scalar-like.

MarcoGorelli: added another commit with get_rows, better to avoid overloads if we can help it

spec/API_specification/dataframe_api/column_object.py
```diff
@@ -1,2 +1,41 @@
...
-    pass
+    def __len__(self) -> int:
+        """
+        Return the number of rows.
+        """
+    
+    def __getitem__(self, row: int) -> object:
+        """
+        Get the element at row index `key`.
+        """
```
kkraus14, line 14: Does typing row as `int` here allow for library specific Scalar objects?

rgommers: It does, as always/everywhere. I'll write up some text about this now.

rgommers: Done in gh-153.

kkraus14, line 14: It still feels a bit funky to me to only support the scalar case via `__getitem__` where I think it would go against folks natural intuition.

MarcoGorelli: there's already `get_rows` for that - perhaps we could start like this, and then we can always overload this one if necessary based on feedback?

jorisvandenbossche: Another use case for getitem (or separate method) is slicing (https://github.com/data-apis/dataframe-api/issues/89 indicates adding a separate method for that as well)

MarcoGorelli:
as discussed yesterday, I've removed `__getitem__` and added `get_value`

for some existing methods (e.g. `Column.mean`), we have `dtype` at the return value to mean "something of the dtype of the column". I've done the same here - orthogonal to this issue, but we should fine a better way of expressing that

kkraus14 (approved):

jorisvandenbossche:
> for some existing methods (e.g. `Column.mean`), we have `dtype` at the return value to mean "something of the dtype of the column". I've done the same here - orthogonal to this issue, but we should fine a better way of expressing that

Should that use `Scalar` instead? (or `Scalar[dtype]` if we want to indicate it is a scalar of the specific dtype)

MarcoGorelli:
yeah I like that more. will address in a separate PR though as there's some unrelated methods which would need changing too, so easier to do them all in one go and stay consistent with this PR

PR #141, opened 2023-04-17
Fill in column a bit

MarcoGorelli:


spec/API_specification/dataframe_api/column_object.py
```diff
@@ -1,2 +1,333 @@
...
+        -------
+        Column
+
+        See also
+        --------
+        isnull
+
+        Notes
+        -----
+        Includes anything with NaN-like semantics, e.g. np.datetime64("NaT").
```
kkraus14, line 331: What is the plan for Pandas with regards to `NaT` and moving to Arrow types and Arrow compute? Arrow doesn't have a concept of `NaT` and there isn't an actual binary value for `NaT` where numpy uses `min(int64)` as a sentinel value for it:
```
In [6]: np.datetime64(np.iinfo(np.int64).min, "ns")
Out[6]: numpy.datetime64('NaT')

In [7]: np.datetime64(np.iinfo(np.int64).min + 1, "ns")
Out[7]: numpy.datetime64('1677-09-21T00:12:43.145224193')
```

MarcoGorelli: that's fine, this is just for anything nan-like, there's no requirement that `NaT` exists

it's the same as DataFrame.isnan

kkraus14: I guess my point is that `NaN` has a binary representation and is a universally agreed upon standard and concept. `NaT` is not and different libraries will have different support, different behaviors, different representations, etc.

Are we saying that supporting the cases of having both `NaT` and `null` values in datetime columns is within the scope of what we're building here?

I.E. we explicitly didn't include `NaT` in the interchange protocol as a null option since in reality it's a sentinel value of `min(int64)`: https://github.com/data-apis/dataframe-api/blob/main/protocol/dataframe_protocol.py#L64-L86

rgommers: I agree with excluding `NaT`, since it's numpy-specific. I don't think that is different here from in the interchange protocol. `Column.isnan` should handle only actual floating-point `nan` values I'd say.

MarcoGorelli: we'll need to update it here too then

https://github.com/data-apis/dataframe-api/blob/86634ea44950b4a49106e4f06bf737aff848d74d/spec/API_specification/dataframe_api/dataframe_object.py#L583-L601

MarcoGorelli: why not leave this implementation-specific and just say that it returns anything nan-like?

rgommers: Isn't this the kind of corner case that is going to lead to differences between implementations when they happen to be using the sentinel value? `NaT` is not a "not-a-number", it's a sentinel value for missing data (I know numpy confused the picture here with its naming choice). So I'd expect it to be dealt with via `isnull` instead.

`nan` is precisely defined, and I don't think there's anything al all that should qualify as "nan-like".

jbrockmendel: Oh here it is.  FWIW the principle i used when including nat here was that it doesnt propagate through comparisons

rgommers: > FWIW the principle i used when including nat here was that it doesnt propagate through comparisons

I'm trying to think about whether that is relevant to including/excluding NaT in `isnan`. I think a simple `col_a > col_b` with `NaT`'s in the same place is going to have problems either way if it doesn't deal with missing values, right? And for dealing with that, one should use `isnull` rather than `isnan`.

I'd think `isnan` isn't useful for anything but floating-point numbers. Although we may not want to reject non-floating-point columns in dataframes for convenience, to avoid making `df.isnan()` hard to use.

Should the semantics be "will only ever return true values for columns with a floating-point dtype, because those are the only ones which can contain NaNs"?

MarcoGorelli: counting nat as null is fine - is practice, there just needs to be a way to exclude invalid values, like
```
mask = ~df.get_column_by_name(label).fill_nan(None).is_null()
df = df.get_rows_by_mask(mask)
```

rgommers (commented):
This looks about ready to merge, thanks Marco! Just a comment on the dtype-related behavior of the reductions.

I did also consider whether reductions should return a size-1 Column rather than a scalar. That may still make sense, but I suspect that returning a scalar value is more practical.

jbrockmendel:
Sum allows timedelta

spec/API_specification/dataframe_api/column_object.py
```diff
@@ -1,2 +1,333 @@
...
+        """
+        Reduction returns a bool.
+
+        Raises
+        ------
+        ValueError
+            If column is not boolean.
+        """
+
+    def min(self, skip_nulls: bool = True) -> float:
```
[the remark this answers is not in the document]
rgommers: Actually, this depends on the method. E.g., `median` should preserve dtype, while `std` turns ints into floats.

rgommers (approved):
Discussed with @MarcoGorelli, we're both happy with the way things look now and the extra details on what reductions return what dtypes. So I'll merge this. This is a large PR, so if there's a detail we missed somewhere, let's deal with that in a follow-up PR.

spec/API_specification/dataframe_api/column_object.py
jorisvandenbossche: > I'm trying to think about whether that is relevant to including/excluding NaT in `isnan`.

I agree it's not relevant for `isnan`, and that it sounds best to keep `isnan` strictly for floating dtypes. 
But I would say it does matter for considering it as null or not. _If_ we specify how a null should behave in comparisons, and _if_ that's to propagate as missing value, then NaT (if you rely purely on the numpy behaviour in your DataFrame library) does not behave as a null. So should then `isnull()` return True?

jorisvandenbossche: So to explicit, you are then OK with `Column.from_sequence([np.NaT], dtype="datetime") > val` and `Column.from_sequence([null], dtype="some_numeric_dtype") > val` to return different results? (and essentially say that the behaviour of nulls is implementation-dependent?)

rgommers: > and essentially say that the behaviour of nulls is implementation-dependent?)

I think we're saying that in a few places already, in particular related to ordering/sorting - which is what your `> val` example seems to get at? Isn't the end result in both cases expected to be the same though? Both cases compare "missing" to a value, so I'd expect a similar outcome.

jorisvandenbossche: > in particular related to ordering/sorting

No, we said that this is implementation-defined _for NaNs_, not for nulls (for nulls the keyword controls the behaviour):

https://github.com/data-apis/dataframe-api/blob/705908137c6421a602765f53a3fbc6ae301901a9/spec/API_specification/dataframe_api/dataframe_object.py#L268-L272

> which is what your `> val` example seems to get at?

For NaNs this is related to sorting, but not for nulls I think. At least if we specify that `null > val` always returns `null`, but so the question is whether we want to specify this, or whether we want to say that behaviour of null is implementation-defined.

>  Isn't the end result in both cases expected to be the same though? Both cases compare "missing" to a value, so I'd expect a similar outcome.

Yes, I would also expect the same result if both are considered as a null. But so that was my original question: NaT does _not_ give the same result:

```
In [19]: pd.NaT > pd.Timestamp("2012-01-01")
Out[19]: False

In [20]: pd.NA > pd.Timestamp("2012-01-01")
Out[20]: <NA>
```

So either it should not be considered as null, or implementations should hide that behaviour of NaT? (or we say this is not covered by the spec and implementation defined)

To make it a bit more concrete:

```
s1 = pd.Series([pd.NaT, "2012-01-01"], dtype="datetime64[ns]")
s2 = pd.Series([pd.NA, 2], dtype="Int64")

In [14]: s1
Out[14]: 
0          NaT
1   2012-01-01
dtype: datetime64[ns]

In [15]: s2
Out[15]: 
0    <NA>
1       2
dtype: Int64

In [17]: s1 >= s1[1]
Out[17]: 
0    False
1     True
dtype: bool

In [18]: s2 >= s2[1]
Out[18]: 
0    <NA>
1    True
dtype: boolean
```

That's with current pandas, where we have a mix of dtypes that are "nullable" (in the sense of following the behaviour of pd.NA), and dtypes that are not (that have the behaviour of NaN / NaT). 
So the question for a spec-compliant DataFrame implementation: which behaviour do we want/require? Is both fine? (and so leave this to the implementation) Or should a spec-compliant implementation hide this different behaviour of NaT compared to NA? (eg an implementation could still use NaT as a missing value sentinel under the hood, but in certain operation have custom handling for it to ensure it behaves spec-compliant)

PR #145, opened 2023-04-19
Fix the Circle CI preview link on PRs

rgommers:
The added token has to be a personal token; I generated that within the CircleCI app.


rgommers:
This will only start working once merged, so I'll do that. It's tested on other repos, so pretty sure this will fix it.

PR #148, opened 2023-04-19
add Column.from_sequence

### LINKED ISSUE #121: Column constructor
For constructing DataFrames, we have `DataFrame.from_dict`

How about constructing a Column? Example of where this would be needed:

https://github.com/mwaskom/seaborn/blob/5d9f37159bbd3ac44c8c8a06825583ba25648525/seaborn/_oldcore.py#L1217-L1220

```python
        for var in axis_variables:
            other_var = {"x": "y", "y": "x"}[var]

            converter = pd.Series(index=self.plot_data.index, name=var, dtype=object)
```

MarcoGorelli:
closes #121

### DIFF
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -1,2 +1,23 @@
+from __future__ import annotations
+
+from typing import Sequence
+
+from ._types import dtype
+
 class Column:
-    pass
+    @classmethod
+    def from_sequence(cls, sequence: Sequence[object], dtype: dtype) -> Column:
+        """
+        Construct Column from sequence of elements.
+
+        Parameters
+        ----------
+        sequence : Sequence[object]
+            Sequence of elements.
+        dtype : str
+            Dtype of result. Must be specified.
+        
+        Returns
+        -------
+        Column
+        """
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -13,7 +13,9 @@ class Column:
         Parameters
         ----------
         sequence : Sequence[object]
-            Sequence of elements.
+            Sequence of elements. Each element must be of the specified
+            ``dtype``, the corresponding Python builtin scalar type, or
+            coercible to that Python scalar type.
         dtype : str
             Dtype of result. Must be specified.
         


rgommers (commented):
Thanks Marco, this looks good. The one important thing that I think is missing is this (from gh-121): "Each element of input sequences must be of that dtype."

I believe the idea was that it should be the relevant Python scalar type, or something that duck types that. So for any integer type, it's `int` (or an object that has `__int__`), for datetimes, `datetime.datetime` instances, etc.

rgommers:
> (or an object that has `__int__`)

and no need to describe that, I'm doing that in a separate page, so it can be linked to.

spec/API_specification/dataframe_api/column_object.py
```diff
@@ -1,2 +1,23 @@
+from __future__ import annotations
+
+from typing import Sequence
+
+from ._types import dtype
+
 class Column:
-    pass
+    @classmethod
+    def from_sequence(cls, sequence: Sequence[object], dtype: dtype) -> Column:
```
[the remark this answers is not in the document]
kkraus14: This is a class method so it can't be `self` I think?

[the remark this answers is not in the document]
rgommers: Ah yes, sorry for the noise. Never use these much, so I forgot. Resolving.

rgommers:
> The one important thing that I think is missing is this (from [gh-121](https://github.com/data-apis/dataframe-api/issues/121)): "Each element of input sequences must be of that dtype."

I guess if `dtype` is explicitly specified, this is currently already implicit if you append "or coercible to that dtype".

spec/API_specification/dataframe_api/column_object.py
```diff
@@ -1,2 +1,25 @@
...
+        Construct Column from sequence of elements.
+
+        Parameters
+        ----------
+        sequence : Sequence[object]
+            Sequence of elements. Each element must be of the specified
+            ``dtype``, the corresponding Python builtin scalar type, or
+            coercible to that Python scalar type.
+        dtype : str
+            Dtype of result. Must be specified.
```
jorisvandenbossche, line 20: Should we start listing the possible values for the strings to specify dtypes? (for example, we don't necessarily want to support all variations of the same dtype that numpy supports?)

rgommers: Ohhhh that is a good catch - this is the very first `dtype` keyword. There's multiple ways we could go there, this is important enough to open a separate issue about. Let me do so now.

rgommers: done in gh-155

PR #149, opened 2023-04-19
remove 'nan-like' from isnan description

MarcoGorelli:


spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -578,12 +578,14 @@ def isnull(self) -> DataFrame:
         Notes
         -----
         Does *not* include NaN-like entries.
+        May optionally include 'NaT' values (if present in an implementation),
+        but note that the Standard makes no guarantees about them.
```
jbrockmendel, line 582: I've lost track of where the discussion is happening for this.  At any rate, we included NaT in the other one originally because it does not propagate through comparisons, which seemed like a reasonable way to split these.

Also what about None?

MarcoGorelli: it's here https://github.com/data-apis/dataframe-api/pull/141

I think that would also count as 'missing' (so, isnull rather than isnan)

rgommers: Let's keep the discussion in gh-141 before merging this (or not).

rgommers (approved):
LGTM, thanks @MarcoGorelli

PR #153, opened 2023-04-26
Add design topic page on use of Python builtin types

rgommers:
This addresses a concern that has come up a number of times about whether it's okay to implement a library-specific object instead of a builtin type. E.g.,
https://github.com/data-apis/dataframe-api/pull/140#discussion_r1171567591

### DIFF
--- a/spec/design_topics/python_builtin_types.md
+++ b/spec/design_topics/python_builtin_types.md
@@ -18,7 +18,7 @@ class DataFrame:
         ...
 
 class Column:
-    def mean(self, skipna: bool = True) -> float:
+    def mean(self, skip_nulls: bool = True) -> float:
         ...
 
 larger = df2 > df1.get_column_by_name('foo').mean()
--- a/spec/design_topics/python_builtin_types.md
+++ b/spec/design_topics/python_builtin_types.md
@@ -1,0 +1,41 @@
+# Python builtin types and duck typing
+
+Use of Python's builtin types - `bool`, `int`, `float`, `str`, `dict`, `list`,
+`tuple`, `datetime.datetime`, etc. - is often natural and convenient. However,
+it is also potentially problematic when trying to write performant dataframe
+library code or supporting devices other than CPU.
+
+This standard specifies the use of Python types in quite a few places, and uses
+them as type annotations. As a concrete example, consider the `mean` method and
+the `float` it is documented to return, in combination with the `__gt__` method
+(i.e., the `>` operator) on the dataframe:
+
+```python
+class DataFrame:
+    def __gt__(self, other: DataFrame | Scalar) -> DataFrame:
+        ...
+    def get_column_by_name(self, name: str, /) -> Column:
+        ...
+
+class Column:
+    def mean(self, skipna: bool = True) -> float:
+        ...
+
+larger = df2 > df1.get_column_by_name('foo').mean()
+```
+
+For a GPU dataframe library, it is desirable for all data to reside on the GPU,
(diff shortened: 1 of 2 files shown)


kkraus14 (approved):
Thanks @rgommers this looks like a good start.

I suspect we'll want to more rigidly define guaranteed functions available on scalar objects in the future, but we can tackle things as they come up.

rgommers:
> I suspect we'll want to more rigidly define guaranteed functions available on scalar objects in the future, but we can tackle things as they come up.

Agreed. I thought about exhaustively documenting the methods each duck type should have, but that's a large/tedious job with limited value for the moment.

spec/design_topics/python_builtin_types.md
```diff
@@ -0,0 +1,41 @@
...
+
+```python
+class DataFrame:
+    def __gt__(self, other: DataFrame | Scalar) -> DataFrame:
+        ...
+    def get_column_by_name(self, name: str, /) -> Column:
+        ...
+
+class Column:
+    def mean(self, skipna: bool = True) -> float:
```
MarcoGorelli, line 21: skipna -> skip_nulls

MarcoGorelli (approved):

spec/design_topics/python_builtin_types.md
```diff
@@ -0,0 +1,41 @@
...
+        ...
+
+larger = df2 > df1.get_column_by_name('foo').mean()
+```
+
+For a GPU dataframe library, it is desirable for all data to reside on the GPU,
+and not incur a performance penalty from synchronizing instances of Python
+builtin types to CPU. In the above example, the `.mean()` call returns a
+`float`. It is likely beneficial though to implement this as a library-specific
+scalar object which duck types with `float`. This means that it should (a) have
```
jorisvandenbossche, line 31: Out of curiosity, cudf doesn't actually do this, right now, is that correct? (for example it does return numpy scalars for numeric types, i.e. what pandas does) 

But cudf would _like_ to do this? (I seem to remember discussions in the past about Scalar objects)

rgommers: I thought it did. If not, maybe @kkraus14 or @shwina can suggest a better example for where cuDF uses scalars.

That said, numpy scalars are also an example of special objects that duck type, they're not Python builtin `float`, `int`, etc.

rgommers:
> For boolean/numeric scalars with an actual value, that's kind of obvious, but how should that behave for a null scalar of those types?

`False` I imagine? It's missing ~= empty. So I'd think it would work like an empty sequence:
```python
>>> bool([])
False
```

jorisvandenbossche:
Empty and missing is not necessarily the same (we currently don't cover nested types, but for example in pyarrow/cudf's list type, a list scalar can be empty or null, which are two separate states) 

Pandas (`bool(pd.NA)`), but that also got some pushback.

rgommers:
So was there an outcome or was it unresolved? I think you only have two options right, `False` or raising an exception. `True` doesn't seem very reasonable.

jorisvandenbossche:
Not really a clear resolution in general (but that's mostly because of my preference to just stay out of this for pyarrow Scalars altogether, and not start with trying to make them behave like python scalars, to avoid all those questions), although we will probably raise for boolean null scalar, just to avoid that people rely on it returning False (that's never the way you should check that it is null)

rgommers:
> Maybe best to include with a `null` object what `bool(null)` does (either raise, or is undefined).

See gh-157 for adding a `null` object.

Issue #115, opened 2023-03-20
How to enable the Standard?


MarcoGorelli:
Probably the most fundamental question

Say we have
- `dfpd`, which is a pandas DataFrame
- `PandasStandardDataFrame`, which is the pandas implementation of the Standard. It takes a pandas DataFrame and returns a class which only has the methods supported by the Standard. Say this is available on PyPI as `pandas-standard`

(...and similarly for modin / cudf / vaex / anyone else)

Right. Say I want to write a function which can accept any DataFrame, like
```python
def clean_column_names(df: DataFrame) -> DataFrame:
    df_standard = <get the relevant Standard implementation>
    mapping = {}
    for column in df_standard.get_column_names():
        mapping[column] = column.lower()
    df_standard = df_standard.rename(mapping)
    return df_standard.dataframe
```

How do we do the first line, i.e. getting `df_standard`?

I think the ideal place to get to would be
```python
df_standard = df.__dataframe_standard__()
```

but this can't happen overnight, especially if we want to stick to the mantra @jbrockmendel had mentioned [here](https://github.com/data-apis/dataframe-api/issues/108#issuecomment-1468878516)

> id like to find an alternative that fits with the "assume pandas changes nothing" mantra

### Phase 1

This a hacky, but allows for quick experimentation without needing to depend on pandas' approvals, or on the pandas' relatively slow release cycle, or those of any other library.

Consumers of the Standard would need to write something like
```python
def enable_standard(df: DataFrame):
    if type(df).split('.')[0] == 'pandas':
        from pandas_standard import PandasStandardDataFrame  # TODO raise if not installed
        return PandasStandardDataFrame(df)
    # and similarly for any other package which might not
    # want to introduce `__dataframe_standard__` right away
    try:
        return df.__dataframe_standard__()
    except AttributeError:
        raise TypeError(f'Expected DataFrame Standard compliant DataFrame, got {type(df)}')
```

At the moment the Consortium is still relatively small, so enumerating the options in `if-then` statements should be manageable.

### Phase 2

Once we've seen that some libraries are actually able to use it to write portable code, then pandas could add a method like
```python
def __dataframe_standard__(self):
    import_optional_dependency("pandas_standard")  # this will raise if not installed
    from pandas_standard import PandasStandardDataFrame
    return PandasStandardDataFrame(self)
```
, and similarly for other libraries.

[...]

Issue #79, opened 2022-06-30
How to make a future dataframe API available?


rgommers:
_This question got asked recently by @mmccarty (and others have brought it up before), so it's worth taking a stab at an answer. Note that this is slightly speculative, given that we only have fragments of a dataframe API rather than a mostly complete syntax + semantics._

A future API, or individual design elements of it, will certainly have (a) new API surface, and (b) backwards-incompatible changes compared to what dataframe libraries already implement. So how should it be made available?

Options include:
1. In a separate namespace, ala `.array_api` in NumPy/CuPy,
2. In a separate _retrievable-only_ namespace, ala `__array_namespace__`,
3. Behind an environment variable (NumPy has done this a couple of times, for example with `__array_function__` and more recently with dtype casting rules changes),
4. With a context manager,
5. With a `from __future__ import new_behavior` type import (i.e., new features on a per-module basis),
6. As an external package, which may for example monkeypatch internals (_added for completeness, not preferred_),

One important difference between arrays and dataframes is that for the former we only have to think about functions, for the latter we're dealing with methods on the main dataframe objects. Hiding/unhiding methods is a little more tricky of course - can be done based on an environment variable set at import time, but it's more annoying with a context manager.

For behavior it's kind of the opposite: likely not all code will work with new behavior, so granular control helps, and a context manager is probably better.

### Experiences with a separate namespace for the array API standard

The short summary of this is:
- there's a problem where we now have two array objects, and supporting both in a code base is cumbersome and requires bi-directional conversions.
- a summary of this problem and approaches taken in scikit-learn and SciPy to work around it are described in https://github.com/data-apis/array-api/issues/400
- in NumPy the preferred solution direction longer term is to make the main `numpy` namespace converge to the array API standard; this takes time because of backwards compatibility constraints, but will avoid the "double namespaces" problem and have multiple other benefits, for example solving long-standing issues that Numba, CuPy etc. are running into.

Therefore, using a separate namespace to implement dataframe API standard features/compatibility should likely not be the preferred solution.

### Using a context manager

Pandas already has a context manager, namely [`pandas.option_context`](https://pandas.pydata.org/docs/reference/api/pandas.option_context.html). This is used for existing options, see `pd.describe_option()`. While most features are related to display, styling and I/O, some features that can be controlled are quite large and similar in style to what we'd expect to see in a dataframe API standard. Examples:
- `mode.chained_assignment` (raise, warn, or ignore)
[...]

PR #156, opened 2023-04-27
Add a way to retrieve the standard-compliant namespace, fill "how to adopt" and "future evolution"

rgommers:
All this is taken over from the array API standard, where the approach is now stable and the kinks have been worked out.

Closes gh-115
Closes gh-79

### DIFF
--- a/spec/purpose_and_scope.md
+++ b/spec/purpose_and_scope.md
@@ -239,17 +239,123 @@ sugar required for fast analysis of data.
 
 ## How to read this document
 
+The API specification itself can be found under {ref}`api-specification`.
+
+For guidance on how to read and understand the type annotations included in
+this specification, consult the Python
+[documentation](https://docs.python.org/3/library/typing.html).
 
 
-
+(how-to-adopt-this-api)=
 ## How to adopt this API
 
+Most (all) existing dataframe libraries will find something in this API standard
+that is incompatible with a current implementation, and that they cannot
+change due to backwards compatibility concerns. Therefore we expect that each
+of those libraries will want to offer a standard-compliant API in a _new
+namespace_. The question then becomes: how does a user access this namespace?
+
+The simplest method is: document the import to use to directly access the
+namespace (e.g. `import package_name.dataframe_api`). This has two issues
+though:
+
+1. Dataframe-consuming libraries that want to support multiple dataframe
+   libraries then have to explicitly import each library.
+2. It is difficult to _version_ the dataframe API standard implementation (see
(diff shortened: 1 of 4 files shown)


kkraus14 (approved):

MarcoGorelli (approved):
Looks fine

to be able to define `concat` as a function, rather than as a dataframe method, could we also add something like a `spec/API_specification/dataframe_api/namespace.py` file in which to define functions such as `concat`?

rgommers:
> to be able to define `concat` as a function, rather than as a dataframe method, could we also add something like a `spec/API_specification/dataframe_api/namespace.py` file in which to define functions such as `concat`?

You can add content to `__init__.py` and list it in `index.rst`, like I did for `null` in gh-157. That should be simple enough I think; I suggest keeping it like that unless we are accumulating an unexpectedly large number of objects in the top-level namespace.

rgommers:
This got two approvals and no unaddressed comments, so I'll hit the green button here.

spec/purpose_and_scope.md
```diff
@@ -239,17 +239,123 @@ sugar required for fast analysis of data.
...
+including additional functionality is not recommended as doing so may hinder
+portability and inter-operation of dataframe libraries within user code.
+
+### Checking a dataframe object for Compliance
+
+Dataframe-consuming libraries are likely to want a mechanism for determining
+whether a provided dataframe is specification compliant. The recommended
+approach to check for compliance is by checking whether a dataframe object has
+an `__dataframe_namespace__` attribute, as this is the one distinguishing
+feature of a dataframe-compliant object.
```
jorisvandenbossche, line 293: In https://github.com/data-apis/dataframe-api/issues/115#issuecomment-1478692817 you mentioned having this attribute on a non-compliant DataFrame object, as a way to go from that to a compliant one (unless I misunderstood the comment). 

In general, this pathway seems missing in the current description? Assume you have written a function to work with a spec compliant dataframe API, is it then up to the user of your library to only pass such dataframes, or do we want a way that the function can check for any dataframe input if it can give in a compliant version of itself?

rgommers: Great catch, thanks Joris. We already have [DataFrame.dataframe](file:///home/rgommers/code/consortium/dataframe-api/spec/_build/html/API_specification/dataframe_object.html#dataframe_api.DataFrame.dataframe) to get back out the user-facing/non-compliant one, but we indeed do need to add the conversion the other way.

Let me open a follow-up issue right now, since I won't get to fixing that up today.

PR #157, opened 2023-04-27
Add `null` object, and update top-level API specification

rgommers:
Also includes `__dataframe_api_version__`, which was already present, into the API specification.

The table of top-level objects/functions will remain pretty sparse, but there's probably a few more things that need adding. `concat` for one, see gh-137.

I'll note that I briefly considered whether this should be spelled `NULL` or `Null`, but went with `null` because that seemed most natural (e.g., `DataFrame.from_sequence([1, 2, null, np.nan]`).

### DIFF
--- a/spec/API_specification/dataframe_api/__init__.py
+++ b/spec/API_specification/dataframe_api/__init__.py
@@ -88,7 +88,7 @@ class null:
     TypeError
         From ``__eq__`` and from ``__bool__``.
 
-        For ``_eq__``: a missing value must not be compared for equality
+        For ``__eq__``: a missing value must not be compared for equality
         directly. Instead, use `DataFrame.isnull` or `Column.isnull` to check
         for presence of missing values.
 
--- a/spec/API_specification/dataframe_api/__init__.py
+++ b/spec/API_specification/dataframe_api/__init__.py
@@ -14,8 +14,9 @@ from ._types import DType
 
 __dataframe_api_version__: str = "YYYY.MM"
 """
-String representing the version of the DataFrame API specification to which the
-conforming implementation adheres.
+String representing the version of the DataFrame API specification to which
+the conforming implementation adheres. Set to a concrete value for a stable
+implementation of the dataframe API standard.
 """
 
 def concat(dataframes: Sequence[DataFrame]) -> DataFrame:
@@ -73,3 +74,36 @@ def dataframe_from_dict(data: Mapping[str, Column]) -> DataFrame:
     DataFrame
     """
     ...
+
+class null:
+    """
+    A `null` singleton object to represent missing data.
+
+    ``null`` may be used when constructing a `Column` from a Python sequence.
+    It supports ``is``, and does not support ``==`` and ``bool``.
+
+    Methods
+    -------
+    __bool__
+    __eq__
+
+    """
+    def __eq__(self):
+        """
+        Raises
+        ------
+        RuntimeError
+            A missing value must not be compared for equality. Use ``is`` to check
+            if an object *is* this ``null`` object, and `DataFrame.isnull` or
+            `Column.isnull` to check for presence of missing values.
+        """
+        ...
+
+    def __bool__(self):
+        """
+        Raises
+        ------
+        TypeError
+            Truthiness of a missing value is ambiguous
+        """
+        ...
--- a/spec/API_specification/dataframe_api/__init__.py
+++ b/spec/API_specification/dataframe_api/__init__.py
@@ -3,7 +3,7 @@ Function stubs and API documentation for the DataFrame API standard.
 """
 from __future__ import annotations
 
-from typing import Mapping, Sequence
+from typing import Mapping, Sequence, Any
 
 from .column_object import *
 from .dataframe_object import *
@@ -102,3 +102,20 @@ class null:
 
     """
     ...
+
+def isnull(value: Any, /) -> bool:
+    """
+    Check if an object is a `null` scalar.
+
+    Parameters
+    ----------
+    value : Any
+        Any input type is valid.
+
+    Returns
+    -------
+    bool
+        True if the input is a `null` object from the same library which
+        implements the dataframe API standard, False otherwise.
+
+    """
--- a/spec/API_specification/dataframe_api/__init__.py
+++ b/spec/API_specification/dataframe_api/__init__.py
@@ -77,33 +77,28 @@ def dataframe_from_dict(data: Mapping[str, Column]) -> DataFrame:
 
 class null:
     """
-    A `null` singleton object to represent missing data.
+    A `null` object to represent missing data.
 
-    ``null`` may be used when constructing a `Column` from a Python sequence.
-    It supports ``is``, and does not support ``==`` and ``bool``.
+    ``null`` is a scalar, and may be used when constructing a `Column` from a
+    Python sequence with `column_from_sequence`. It does not support ``is``,
+    ``==`` or ``bool``.
 
-    Methods
-    -------
-    __bool__
-    __eq__
+    Raises
+    ------
+    TypeError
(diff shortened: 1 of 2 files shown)


spec/API_specification/dataframe_api/__init__.py
```diff
@@ -9,6 +9,41 @@
...
+implementation of the dataframe API standard.
 """
+
+
+class null:
+    """
+    A `null` singleton object to represent missing data.
+
+    ``null`` may be used when constructing a `Column` from a Python sequence.
+    It supports ``is``, and does not support ``==`` and ``bool``.
```
kkraus14, line 23: I think this could use some design discussion. Some initial questions that come to mind:

- We don't currently have a `NULL` type, would this object then be untyped?
  - PyArrow's singleton for example has a `NULL` type
  - Pandas's singleton does not have a type as far as I can tell
- If someone does something like calculate the max of a column with nulls in it with `skip_nulls=False`, would it be expected to return this singleton?
  - That would lose the typing information on that null scalar
  - That would also require synchronizing in order to introspect if the result is null

rgommers: > We don't currently have a `NULL` type, would this object then be untyped?

I'm not sure that this is any different from other classes, like `DataFrame`. They have a name, but no type beyond that. To type this sensible in a way that helps downstream libraries interact with multiple implementors, it'd need a `typing.Protocol`, an ABC, or some such thing.

Personally I am happy to punt on typing, and only require here that `null` exists under that name and behaves as specified.

> If someone does something like calculate the max of a column with nulls in it with `skip_nulls=False`, would it be expected to return this singleton?

Yes, I think so. I don't think there's anything else we could return, right?

rgommers: > That would also require synchronizing in order to introspect if the result is null

Perhaps not - I think this is the same as methods that return `float`. You can duck type scalars to keep them on the GPU.

jorisvandenbossche: > Personally I am happy to punt on typing, and only require here that null exists under that name and behaves as specified.

I think Keith was not talking about "typing" in the sense of python type annotations, but rather of data types (dtype). For example, if you do `Column.from_sequence([null])`, what would be the type of that Column? 
(the answer here might also be: you should specify the `dtype` instead of letting it infer)

> Yes, I think so. I don't think there's anything else we could return, right?

A specific implementation could return their own Scalar object of the specific type (which can be a null scalar), instead of this generic null object

rgommers: > the answer here might also be: you should specify the dtype instead of letting it infer

yes, I think that's the answer. and `dtype` is mandatory in `from_sequence` for that reason. we can document "no dtype inference anywhere" as a design rule separately?

> A specific implementation could return their own Scalar object of the specific type (which can be a null scalar), instead of this generic null object

Agreed, and that's always allowed for any scalar type (see https://data-apis.org/dataframe-api/draft/design_topics/python_builtin_types.html). I meant semantically there's nothing else one can return. A duck-typed `null` is still `null`.

kkraus14: The docstring here implied using `is` to check if a scalar is null which would imply that this is a singleton and if we wanted a ducktyped null scalar to be a singleton it would require synchronization.

MarcoGorelli (approved):
looks fine to me

spec/API_specification/dataframe_api/__init__.py
rgommers: That's a good point, let me think about how to rephrase this so that duck typing works, and explicitly reference the docs page on that topic.

rgommers: I rephrased this, forbidding `is null` and explicitly addressing that `null` may be duck typed so it can live on device.

spec/API_specification/dataframe_api/__init__.py
```diff
@@ -9,6 +9,36 @@
...
+    ``==`` or ``bool``.
+
+    Raises
+    ------
+    TypeError
+        From ``__eq__`` and from ``__bool__``.
+
+        For ``_eq__``: a missing value must not be compared for equality
+        directly. Instead, use `DataFrame.isnull` or `Column.isnull` to check
+        for presence of missing values.
```
kkraus14, line 33: How can one check if a scalar is null reliably then? Do we need a namespace `isnull` method?

I.E. someone runs `my_column.sum(skip_null=False)` and it yields `null` as the result.

rgommers: Good question, I see a few options:
- Use `is`. Runs into the "that does not allow duck typing" problem
- Use `==`. This may also not work when duck typing, or at least not desirable - `null_duck_on device == null` is not going to do what you want if `null` lives on host, and also it may cause folks to do `column == null` which probably works if one implements `__eq__`
- Add a free `isnull` function. Better perhaps, only cost is a new object in the namespace.
- Add a comparison method, e.g. `null.is_equal(val)`. Doesn't read all that well.

jorisvandenbossche: I don't think the first two options are viable: `is` doesn't work for type specific nulls (and I think we shouldn't disallow that if an implementation uses that?), and `==` shouldn't give the behaviour you want here (it shouldn't necessarily return True, just like nan == nan also doesn't give true)

A free `is_null` function might be the easiest? (a method on the generic namespace.null method probably works as well, but a top-level function feels a bit more natural to me)

rgommers: Added a free `isnull` function which checks for a scalar `null`. That should address this issue.

kkraus14 (approved):

spec/API_specification/dataframe_api/__init__.py
```diff
@@ -73,3 +74,48 @@ def dataframe_from_dict(data: Mapping[str, Column]) -> DataFrame:
...
+    Notes
+    -----
+    Like for Python scalars, the ``null`` object may be duck typed so it can
+    reside on (e.g.) a GPU. Hence, the builtin ``is`` keyword should not be
+    used to check if an object *is* the ``null`` object.
+
+    """
+    ...
+
+def isnull(value: Any, /) -> bool:
```
MarcoGorelli, line 106: should we use `: object` instead? `Any` turns off the typechecker, and if we're saying that any input type is valid, then we probably want to refrain from using any type-specific methods on `value`?

spec/API_specification/dataframe_api/__init__.py
```diff
@@ -73,3 +74,48 @@ def dataframe_from_dict(data: Mapping[str, Column]) -> DataFrame:
...
+    ``null`` is a scalar, and may be used when constructing a `Column` from a
+    Python sequence with `column_from_sequence`. It does not support ``is``,
+    ``==`` or ``bool``.
+
+    Raises
+    ------
+    TypeError
+        From ``__eq__`` and from ``__bool__``.
+
+        For ``_eq__``: a missing value must not be compared for equality
```
MarcoGorelli, line 91:
```suggestion
        For ``__eq__``: a missing value must not be compared for equality
```

MarcoGorelli (approved):
couple of minor (non-blocking) comments, rest looks good

spec/API_specification/dataframe_api/__init__.py
MarcoGorelli: can always address this later though - I'd say let's just get this in now

rgommers: > and if we're saying that any input type is valid, then we probably want to refrain from using any type-specific methods on `value`?

Sure. There's nothing saying that those exist or must be used, right? This seems like a detail that an implementer has to get right.

PR #160, opened 2023-05-02
add Column.is_in

### LINKED ISSUE #159: Column.is_in
In the last call it was brought up that perhaps we should add `Column.is_in`

This can be achieved with the existing Standard methods:
```python
result = column == values[0]
for value in values[1:]:
    if math.isnan(value):
        result |= column.isnan()
    elif value is namespace.null:
        result |= column.isnull()
    else:
        result |= column == value
```
so it really just be a convenience method. No objections to adding it anyway

MarcoGorelli:
closes #159

### DIFF
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -439,7 +439,7 @@ class Column:
         ----------
         values : Column
             Contains values to compare against. May include ``float('nan')`` and
-            ``namespace.null``, which which case ``'nan'`` and ``null`` will
+            ``null``, which case ``'nan'`` and ``null`` will
             respectively return ``True`` even though ``float('nan') == float('nan')``
             isn't ``True``.
             The dtype of `values` must match the current column's dtype.
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -396,6 +396,7 @@ class Column:
             ``namespace.null``, which which case ``'nan'`` and ``null`` will
             respectively return ``True`` even though ``float('nan') == float('nan')``
             isn't ``True``.
+            The dtype of `values` must match the current column's dtype.
 
         Returns
         -------
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -385,15 +385,17 @@ class Column:
         In particular, does not check for `np.timedelta64('NaT')`.
         """
 
-    def is_in(self, values: Sequence[object]) -> Column[bool]:
+    def is_in(self, values: Column) -> Column[bool]:
         """
         Indicate whether the value at each row matches any value in `values`.
 
         Parameters
         ----------
-        values : Sequence[object]
-            Values to compare against. May include ``float('nan')`` and
-            ``namespace.null``.
+        values : Column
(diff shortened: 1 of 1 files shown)


spec/API_specification/dataframe_api/column_object.py
```diff
@@ -384,3 +384,19 @@ def isnan(self) -> Column:
...
         In particular, does not check for `np.timedelta64('NaT')`.
         """
+
+    def is_in(self, values: Sequence[object]) -> Column[bool]:
+        """
+        Indicate whether the value at each row matches any value in `values`.
+
+        Parameters
+        ----------
+        values : Sequence[object]
```
jorisvandenbossche, line 394: I suppose we can be stricter here, and require that this is a Column object?

spec/API_specification/dataframe_api/column_object.py
```diff
@@ -384,3 +384,19 @@ def isnan(self) -> Column:
...
+            Values to compare against. May include ``float('nan')`` and
+            ``namespace.null``.
```
jorisvandenbossche, line 396: We might need to specify the behaviour for NaN and null (that those give a "match", i.e. resulting in a True value, in contrast to standard equality with `==`)

spec/API_specification/dataframe_api/column_object.py
MarcoGorelli: yeah sounds fine, thanks!

kkraus14 (approved):

spec/API_specification/dataframe_api/column_object.py
```diff
@@ -384,3 +384,21 @@ def isnan(self) -> Column:
...
         In particular, does not check for `np.timedelta64('NaT')`.
         """
+
+    def is_in(self, values: Column) -> Column[bool]:
+        """
+        Indicate whether the value at each row matches any value in `values`.
+
+        Parameters
+        ----------
+        values : Column
```
jbrockmendel, line 394: should the dtypes match?

MarcoGorelli: yeah probably, else pandas does magic stuff like
```python
In [3]: pd.date_range('2000', periods=3).to_series().isin(['2000-01-01'])
Out[3]:
2000-01-01     True
2000-01-02    False
2000-01-03    False
Freq: D, dtype: bool
```
which other libraries almost certainly won't do

jbrockmendel: btw I think we (pandas) should deprecate that casting

MarcoGorelli: yeah that'd be good

spec/API_specification/dataframe_api/column_object.py
```diff
@@ -384,3 +384,22 @@ def isnan(self) -> Column:
...
+            Contains values to compare against. May include ``float('nan')`` and
+            ``namespace.null``, which which case ``'nan'`` and ``null`` will
```
rgommers, line 396: can you leave off `namespace`? Just plain `null` should do fine here.

spec/API_specification/dataframe_api/column_object.py
rgommers: I'll resolve this - perhaps you want to turn the above "we should deprecate this in pandas" into a pandas issue @jbrockmendel or @MarcoGorelli?

PR #161, opened 2023-05-02
make skip_nulls keyword-only

MarcoGorelli:


rgommers (approved):
Looks like a good idea, thanks! We should remember using keyword-only and positional-only arguments wherever that makes sense I think.

MarcoGorelli:
thanks - merging then, hope it's ok, this is really minor anyway

Issue #139, opened 2023-04-17
to_array


MarcoGorelli:
let's split this off from `to_iterable` (https://github.com/data-apis/dataframe-api/issues/109)

PR #163, opened 2023-05-03
Add a `to_array_obj` method to Column

rgommers:
Closes gh-139

### DIFF
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -17,6 +17,54 @@ class Column:
     constructor functions or an already-created dataframe object retrieved via
 
     """
+    def to_array_obj(self) -> object:
+        """
+        Obtain an object that can be used as input to ``asarray`` or ``from_dlpack``
+
+        The returned object must only be used for one thing: calling the ``asarray``
+        or ``from_dlpack`` functions of a library implementing the array API
+        standard, or equivalent ``asarray``/``from_dlpack`` functions to the
+        one in that standard. In practice that means that the returned object
+        may either already *be* an array type of a specific array library, or
+        it may implement one or more of the following methods or behaviors:
+
+        - ``__dlpack__``
+        - the Python buffer protocol
+        - ``__array__``
+        - ``__array_interface__``
+        - ``__cuda_array_interface__``
+        - the Python ``Sequence`` interface
+        - any other method that is known to work with an ``asarray`` function
+          in a library of interest
+
+        Importantly, the returned object must only implement/expose those
+        methods that work. An ``asarray`` function may try to use any of the
+        above methods, and the order in which it does so is not guaranteed.
+        Hence it is expected that if a method is present, it works correctly
+        and does not raise an exception (e.g., because of an unsupported dtype
+        or device).
+
+        .. admonition:: Tip
+
+           One way to expose methods only if they will work for the dtype,
+           device, and other characteristics of the column is to hide access to
+           the methods dynamically, so ``hasattr`` does the right thing. For
+           example::
+
+              def __dir__(self):
+                  methods = dir(self.__class__)
+                  attrs = list(self.__dict__.keys())
+                  keys = methods + attrs
+                  if not self.dtype in _dlpack_supported_dtypes:
+                      keys.remove("__dlpack__")
+
+                  return keys
+
+              def __dlpack__(self):
+                  ...
+
+        """
+
     @classmethod
     def from_sequence(cls, sequence: Sequence[object], dtype: dtype) -> Column:
         """
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -17,7 +17,7 @@ class Column:
     constructor functions or an already-created dataframe object retrieved via
 
     """
-    def to_array_obj(self) -> object:
+    def to_array_obj(self, *, null_handling: str | None = None) -> object:
         """
         Obtain an object that can be used as input to ``asarray`` or ``from_dlpack``
 
@@ -62,6 +62,38 @@ class Column:
 
               def __dlpack__(self):
                   ...
+
+        Parameters
+        ----------
+        null_handling : str or None
+            Determine how to treat ``null`` values that may be present in the
+            column. Valid options are:
+
+            - ``None`` (default): no special handling. This assumes that either
+              no missing values are present, or there is an array type with
+              native support for missing values that is/can be converted to.
+              *Note: there is currently no such library that is in wide use;
+              NumPy's masked arrays are non-recommended, and other array
+              libraries do not support missing values at all.*
+            - ``raise``: always raise a ``ValueError`` if nulls are present.
+            - ``to-nan``: for floating-point dtypes, convert any nulls to ``nan``.
+              For other dtypes, do the same as ``None``.
+
+            Note that if it is desired to convert nulls to a dtype-specific
+            sentinel value, the user should do this before calling
+            ``is_array_obj`` with `.isnull()` and replacing the values
+            directly.
+
+        Raises
+        ------
+        TypeError
+            In case it is not possible to convert the column to any (known) array
+            library type, or use any of the possible interchange methods.
+            This can be due to the dtype (e.g., no array library supports datetime
+            dtypes with a time zone), device, or other reasons.
+        ValueError
+            If the column contains ``null`` values which prevent returning an
+            array object.
 
         """
 
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -84,6 +84,15 @@ class Column:
             ``is_array_obj`` with `.isnull()` and replacing the values
             directly.
 
+        Returns
+        -------
+        array_obj : object
+            Either a custom object or an instance of the  array type of a
+            specific array library. In the latter case, the array must have the
+            closest matching dtype of the array library (e.g., a column with a
+            floating-point dtype must produce the corresponding floating-point
+            dtype of the same precision that the array library offers).
+
         Raises
         ------
         TypeError


spec/API_specification/dataframe_api/column_object.py
```diff
@@ -17,6 +17,54 @@ class Column:
...
+        - the Python ``Sequence`` interface
+        - any other method that is known to work with an ``asarray`` function
+          in a library of interest
+
+        Importantly, the returned object must only implement/expose those
+        methods that work. An ``asarray`` function may try to use any of the
+        above methods, and the order in which it does so is not guaranteed.
+        Hence it is expected that if a method is present, it works correctly
+        and does not raise an exception (e.g., because of an unsupported dtype
+        or device).
```
jorisvandenbossche, line 45: Is the expected behaviour that calling `to_array_obj()` already raises an error in case of an unsupported dtype?

jorisvandenbossche:
Should we specify the behaviour in case of nulls? (do some automatic conversion, or raise an error?) Or add a keyword that controls to which value it is converted?

spec/API_specification/dataframe_api/column_object.py
rgommers: Good point. I made this raise `TypeError`.

rgommers:
> Should we specify the behaviour in case of nulls? (do some automatic conversion, or raise an error?) Or add a keyword that controls to which value it is converted?

Good question - I added a keyword and specified `ValueError` when `null`s cause a column to not be convertible.

spec/API_specification/dataframe_api/column_object.py
```diff
@@ -63,6 +63,38 @@ def __dir__(self):
...
+            column. Valid options are:
+
+            - ``None`` (default): no special handling. This assumes that either
+              no missing values are present, or there is an array type with
+              native support for missing values that is/can be converted to.
+              *Note: there is currently no such library that is in wide use;
+              NumPy's masked arrays are non-recommended, and other array
+              libraries do not support missing values at all.*
+            - ``raise``: always raise a ``ValueError`` if nulls are present.
+            - ``to-nan``: for floating-point dtypes, convert any nulls to ``nan``.
```
jorisvandenbossche, line 79: This _could_ be generalized to apply to any numeric type, instead of only floating points (i.e. allowing to convert ints with nulls to float with nans)

rgommers: True. I'm not sure how desirable that is, it may be pragmatic in the absence of a good alternative. Let's see what others think about this.

MarcoGorelli (changes requested):
Thanks for doing this

My only request would be to force the target dtype to be specified (to prevent implementers from stuffing everything into an object ndarray - looking at you, pandas nullable dtypes)

rgommers:
> My only request would be to force the target dtype to be specified

That would be counterproductive I think. The intended usage is this:
```python
# If you want say a PyTorch tensor, replace `np.asarray` with `torch.Tensor`:
np.asarray(col.to_array_obj())
```
In, e.g., Matplotlib that is supposed to work for pretty much any dtype. How would you usefully fill in a `dtype=` keyword here?

>  (to prevent implementers from stuffing everything into an object ndarray - looking at you, pandas nullable dtypes)

That is a mis-feature in pandas, and it should be solved in pandas imho. This standard nor any other array library even has the concept of an object array.

If it would help to address this specific concern, I'd rather specify something like "a dataframe library must produce the closest matching dtype of the array library in case it produces an array type instance itself (e.g. _"a column with a floating-point dtype must produce the corresponding floating-point dtype of the same precision that the array library offers"_.

MarcoGorelli:
> I'd rather specify something like "a dataframe library must produce the closest matching dtype of the array library in case it produces an array type instance itself (e.g. "a column with a floating-point dtype must produce the corresponding floating-point dtype of the same precision that the array library offers".

sure, sounds good, that makes it clear that pandas' nullable dtypes `to_numpy` default behaviour is considered non-compliant

jorisvandenbossche:
Yes, see also what Ralf already added about nulls causing a ValueError. Most of the cases that pandas currently gives object dtype would already give an error for that reason.

MarcoGorelli (approved):
This is fine actually, apologies, it already says

> of a library implementing the array API standard

which excludes the object dtype anyway

Looks good to me, thanks!

MarcoGorelli:
Looks good - OK to merge?

spec/API_specification/dataframe_api/column_object.py
rgommers: This option could also be removed completely, assuming we add `fill_null(scalar_value)`, in favor of a separate `col.fill_null(float('nan')).to_array_obj()`.

rgommers: Syntax-wise, keeping `'to-nan'` is probably nicer though.

rgommers:
> Looks good - OK to merge?

I added the language regarding the "may produce object dtype" concern, and am now happy with this PR. There is one open point on what to do with `null_handling='to-nan'` - I don't have a strong preference there. We could discuss over the next day and merge then if everyone is happy.

spec/API_specification/dataframe_api/column_object.py
```diff
@@ -17,6 +17,86 @@ class Column:
     constructor functions or an already-created dataframe object retrieved via
 
     """
+    def to_array_obj(self, *, null_handling: str | None = None) -> object:
+        """
+        Obtain an object that can be used as input to ``asarray`` or ``from_dlpack``
```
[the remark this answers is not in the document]
rgommers: Mostly, yes. The main pain point there is that the array API standard doesn't mandate that `asarray` has to support DLPack. There was a desire to keep that orthogonal (long discussion, and I believe you were involved in that). So now if we have a column with a dtype which can support DLPack, that will work with `numpy.asarray` but not necessarily with other libraries.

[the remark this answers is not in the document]
rgommers: So we can just drop `from_dlpack` from the description and leave everything else unchanged.

spec/API_specification/dataframe_api/column_object.py
```diff
@@ -17,6 +17,86 @@ class Column:
...
+        it may implement one or more of the following methods or behaviors:
+
+        - ``__dlpack__``
+        - the Python buffer protocol
+        - ``__array__``
+        - ``__array_interface__``
+        - ``__cuda_array_interface__``
+        - the Python ``Sequence`` interface
+        - any other method that is known to work with an ``asarray`` function
+          in a library of interest
```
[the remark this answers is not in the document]
rgommers: > Is the goal of this function to work generally with multiple libraries `asarray` implementations or just a specific library's implementation?

I'd say neither is 100% accurate. It's more the former, however it does make sense to add library-specific methods - and in fact, `__array__` is one already. There is no single interchange method that is supported by all libraries' `asarray` function anyway, nor is that a good goal because it would eliminate support for dtypes like string and datetime which numpy supports and some other array library could support, but most libraries never will.

spec/API_specification/dataframe_api/column_object.py
```diff
@@ -17,6 +17,95 @@ class Column:
     constructor functions or an already-created dataframe object retrieved via
```
amueller, line 17: Not related to the PR bu the sentence seems to be missing something?

spec/API_specification/dataframe_api/column_object.py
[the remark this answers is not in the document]
kkraus14: I guess my fear is that say some dataframe library `fancy_dataframe` could expect to work downstream with `fancy_array` and return something that supports `__fancy_array__` which `fancy_array.asarray` handles, but a developer calls `np.asarray` against the output of this and it understandably doesn't work.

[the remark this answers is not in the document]
rgommers: I don't think that fear is justified, because:
- this method already allows returning an instance of `fancy_array` directly anyway,
- the returned object should have other, more standard methods too if possible, so `np.asarray` can still work
- if there's something really specific in the column such that it's not possible to add anything but `__fancy_array__` anway, then nothing is lost by doing so

MarcoGorelli (changes requested):
A suggestion which came out of today's call was to not have `to_array_obj` in the end, but to instead have `Column.column`. And put a big warning in its docs, telling users that if they invoke it, then they are leaving the Standard territory and that we make no guarantees about what will or won't work with it

I'd prefer that TBH. Any objections?

(requesting changes so this doesn't get accidentally merged before we've figured it out)

jorisvandenbossche:
Some other options to consider:

- Give up on trying to be array-library agnostic, and just add an explicit `to_numpy()` so at least we have something that is dataframe-library-independent to get a numpy array.
- Limit ourselves to numeric data types for now, so we could specify that the returned object from `to_array_obj()` needs to have `__dlpack__`, to make this a more well defined specification

kkraus14:
I think the challenge is that the API as it currently exists is too ambiguous and encourages use as opposed to it being an "escape hatch". I think what we generally agreed on the call today was:

- There's desire for an API to return an array that guarantees you can use the array API against it, but doesn't make any guarantees about memory layout
- There's desire for an API to return something that guarantees `__dlpack__` interchange compatibility
- These two shouldn't be the same API
- There's going to need to be an escape hatch to use library specific functionality that is not standardized and this `to_array_obj` is just a single instance of that. We provide an escape hatch at the DataFrame level today via the `.dataframe` property: https://github.com/data-apis/dataframe-api/blob/main/spec/API_specification/dataframe_api/dataframe_object.py#L83-L90. Providing an equivalent of that at the Column level serves the purpose of `to_array_obj` as well as serving the purpose of future cases where there's desire to break away from the standard.

MarcoGorelli:
I noticed that plotly have this:

https://github.com/plotly/plotly.py/blob/4ab63db64b09d7e8ba28ea2eb67fbefe7c18ffd4/packages/python/plotly/_plotly_utils/basevalidators.py#L169-L174

in case it's `False`, then the iterate over the elements 1 by 1

----

If we can find closure to this discussion, then that's probably everything that plotly would need to be DataFrame-agnostic (from my initial experimentation at least)

plotly's dataframe handling is _way_ simpler than seaborn's, so they might be a better target for now (they've also expressed interest)

kkraus14:
> I noticed that plotly have this:
> 
> https://github.com/plotly/plotly.py/blob/4ab63db64b09d7e8ba28ea2eb67fbefe7c18ffd4/packages/python/plotly/_plotly_utils/basevalidators.py#L169-L174
> 
> in case it's `False`, then the iterate over the elements 1 by 1

The link here 404s, any chance you could update it?

MarcoGorelli:
thanks @kkraus14 , have fixed

MarcoGorelli:
superseded by #209 - closing then, thanks all for the discussion

Issue #162, opened 2023-05-02
Inserting a column of nans / namespace.column_class


MarcoGorelli:
Motivation: in seaborn they insert a column of `'nan'`s at one point, which in pandas is as simple as
```python
df[new_column_name] = np.nan
```
How do we do that within the Standard? I'm currently doing something like:
```python
df: DataFrame
namespace = df.__dataframe_namespace__()
df= df.insert(
    df.shape()[1],
    new_column_name,
    namespace.column_class.from_sequence([float('nan')] * len(df)), dtype="float64"),
)
```
, is that OK?

If so, it would require adding a `namespace.column_class` method, which would do something like:
```python
>>> df: PandasDataFrame
>>> namespace = df.__dataframe_namespace__()
>>> namespace.column_class
PandasColumn
>>> df: PolarsDataFrame
>>> namespace = df.__dataframe_namespace__()
>>> namespace.column_class
PolarsColumn
>>> # etc...
```

rgommers:
That looks a bit hard to understand. It suggests a change to `insert` may be useful, as well as a free function constructor, so it can be written as:
```python
ns = df.__dataframe_namespace__()
col_nans = ns.from_sequence([float('nan')] * df.shape()[0]), dtype="float64"),
df= df.insert('new_nancol', col_nans, loc=df.shape()[1])
``` 

This highlights my the uneasiness I felt with the `classmethod` constructors: it enforces knowing the exact name of the class. There is a reason why constructors are functions in array libraries, I think it's strictly superior. Adding a function to retrieve that class name instead can't be the right solution.

MarcoGorelli:
`ns.from_sequence` looks good too, thanks!

Issue #155, opened 2023-04-27
How to specify a dtype with a `dtype=` keyword


rgommers:
`DataFrame.from_sequence` introduces a `dtype=` keyword, and specifies it's a string without going into details. This is a potentially tricky topic. There's a number of ways this could go:

- Indeed use strings, e.g. `dtype='float64'`
- Use canonical names like in the array API, e.g. `dtype=float64`
- Use the same as done in the [interchange protocol](https://data-apis.org/dataframe-protocol/latest/API.html), e.g. `Dtype = Tuple[DtypeKind, int, str, str]`

My first instinct would be to use canonical names rather than strings, but it depends a bit on how often we'll use dtypes I think.

kkraus14:
If we use strings, I would propose we use the Arrow format strings to be consistent with where we use strings elsewhere.

How would canonical names work with things like nullability and categoricals, where presumably we want to capture the category type and index type?

We could also consider doing something like having a Data Type class specification that everyone then duck types.

jorisvandenbossche:
The Arrow C interface format strings (https://arrow.apache.org/docs/dev/format/CDataInterface.html#data-type-description-format-strings) are not particularly meant to be human friendly (descriptive) strings, so I would consider that as a big drawback to use those in actual python code. 

It also doesn't cover categorical data type (in the Arrow C interface, the format string is the index type (eg int32), and the categories' type can be inferred separately from the dictionary array)

rgommers:
Okay, here is a proposal:

- For dtypes supported by the array API standard, we use [the same design](https://data-apis.org/array-api/latest/API_specification/data_types.html), so this gives us a `bool`, `uintX`, `intX` and `floatX` (in the top-level namespace, or in a `dtypes` sub-namespace if preferred). These are basically plain objects with given names, not much more than that.
    - Do we want/need `complex64`/`complex128` too, or not? My assumption is not.
- No string aliases allowed, to stay with a single way of doing things.
- For strings we can do the same, because I don't think there's anything we need beyond a name. So `string`. That'd be a variable-length string with support for nulls.
- Categorical and datetime dtypes are the harder ones. These seem to require classes that can be instantiated:

```python
# Similar to pandas.CategoricalDtype
class CategoricalDtype:
    def __init__(self, categories: Column | None = None, *, ordered=False)
        """
        If categories is None: values-only categorical
        If categories is a Column: dictionary-style categorical
        """

    @property
    def categories(self)

    @property
[...]

PR #164, opened 2023-05-04
move from_sequence to namespace

MarcoGorelli:
closes https://github.com/data-apis/dataframe-api/issues/162

### DIFF
--- a/spec/API_specification/dataframe_api/__init__.py
+++ b/spec/API_specification/dataframe_api/__init__.py
@@ -3,7 +3,7 @@ Function stubs and API documentation for the DataFrame API standard.
 """
 from __future__ import annotations
 
-from typing import Sequence
+from typing import Mapping, Sequence
 
 from .column_object import *
 from .dataframe_object import *
@@ -18,7 +18,7 @@ String representing the version of the DataFrame API specification to which the
 conforming implementation adheres.
 """
 
-def from_sequence(sequence: Sequence[object], dtype: dtype) -> Column:
+def column_from_sequence(sequence: Sequence[object], dtype: dtype) -> Column:
     """
     Construct Column from sequence of elements.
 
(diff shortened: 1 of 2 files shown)


spec/API_specification/dataframe_api/__init__.py
```diff
@@ -1,14 +1,38 @@
...
+from ._types import dtype
+
 
 __dataframe_api_version__: str = "YYYY.MM"
 """
 String representing the version of the DataFrame API specification to which the
 conforming implementation adheres.
 """
+
+def from_sequence(sequence: Sequence[object], dtype: dtype) -> Column:
```
jorisvandenbossche, line 21: Should it then be "column_from_sequence" to make it clear it is constructing a Column and not a DataFrame?

jorisvandenbossche (commented):
Should we do the same for DataFrame.from_dict?

MarcoGorelli:
good one, thanks for your review - have updated

spec/API_specification/dataframe_api/__init__.py
```diff
@@ -1,14 +1,54 @@
...
+from ._types import dtype
+
 
 __dataframe_api_version__: str = "YYYY.MM"
 """
 String representing the version of the DataFrame API specification to which the
 conforming implementation adheres.
 """
+
+def column_from_sequence(sequence: Sequence[object], dtype: dtype) -> Column:
```
rgommers, line 21: consider making `dtype` keyword-only?

rgommers (approved):
This looks about ready to me. The `dtype: str` keyword may still change based on gh-155, but that can be done afterwards to not block this PR or other ones - that needs its own PR to deal with `dtype`'s consistently all in one go.

spec/API_specification/dataframe_api/__init__.py
MarcoGorelli, line 21:
```suggestion
def column_from_sequence(sequence: Sequence[object], *, dtype: dtype) -> Column:
```

MarcoGorelli:
merging then, thanks for reviewing!

PR #166, opened 2023-05-10
add Column.dtype

### LINKED ISSUE #165: Column.dtype
We need such a property, right? Else how do we check that dtypes match?

MarcoGorelli:
closes #165

### DIFF
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -54,6 +54,12 @@ class Column:
         """
         raise NotImplementedError("'__iter__' is intentionally not implemented.")
 
+    @property
+    def dtype(self) -> dtype:
+        """
+        Return data type of column.
+        """
+
     def get_rows(self, indices: Column[int]) -> Column:
         """
         Select a subset of rows, similar to `ndarray.take`.


rgommers:
Strangely enough CI is not happy with this:
```
docstring of dataframe_api.Column.dtype:1: WARNING: py:class reference target not found: dataframe_api._types.dtype
```

jorisvandenbossche:
Should this say _something_ about the returned object? Or do just see it as an opaque object without any guarantees? (except that you can pass it to functions with an argument that expects a dtype?)

Or do we want to specify that certain attributes or methods are available? (eg is `__eq__` expected to work)

rgommers:
> Should this say _something_ about the returned object? Or do just see it as an opaque object without any guarantees?

I'd say there are guarantees for dtypes that are returned, but that's gh-155 rather than this PR. We already use `dtype` in some places, and this PR is just one more - so this PR doesn't have to be the one to work out the hard part.

MarcoGorelli:
> so this PR doesn't have to be the one to work out the hard part.

besides, the hard part of this PR is figuring out how to make it pass CI ;)

rgommers:
> it was just that `def dtype` shadows the `dtype` type. Solutions could either be `import dtype as dtype_t`, or rename `def dtype` to a function like `def get_dtype`. I've gone with the latter

I feel like the name chosen for type annotation earlier is a detail, and should not prevent choosing the most obvious name for the property (or method, but I think a `.dtype` property as you had it is best). 

`dtype_t` works, although perhaps `DType` is better - I think that's the most often used solution, capitalize type annotation names.

rgommers (approved):
This is green and seems as uncontroversial as it can get for an API addition, so let's get this in.

Issue #142, opened 2023-04-18
nan to null strategy?


MarcoGorelli:
Both cudf and polars have both null and nan (and there's some discussion about having in pandas too), and they both have methods for converting nans to nulls:

- cudf: `DataFrame.nans_to_nulls`: https://docs.rapids.ai/api/cudf/stable/api_docs/api/cudf.dataframe.nans_to_nulls
- polars: `.fill_nan(None)`

I think we need something similar here. Perhaps `fill_nans` and `fill_nulls` would be most generic?

rgommers:
That seems like useful/necessary functionality indeed.

> Perhaps `fill_nans` and `fill_nulls` would be most generic?

Those both seems useful in principle. That's for actually filling values though (see the [polars `fill_nan` doc](https://pola-rs.github.io/polars/py-polars/html/reference/dataframe/api/polars.DataFrame.fill_nan.html)), so they seem to be orthogonal to `nans_to_nulls` (it's not clear that polars allows `fill_nan(null)`).

I also note that in polars, `fill_nan` is a much simpler API than `fill_null`, the latter has a bunch of strategies that are potentially complex to implement. cuDF's [`fillna`](https://docs.rapids.ai/api/cudf/stable/api_docs/api/cudf.dataframe.fillna) has different strategies.

Given that `fill_null` is complex, maybe start with just dealing with converting `nan`'s into (a) nulls, and (b) scalar values?

kkraus14:
`nans_to_nulls` could be represented as `fill_nan(null)` more flexibly. Is there an API for a more general scalar value replacement? I.E. replace value `a` with value `b` in a Column? If so would probably make more sense to ensure that both `null` and `NaN` are supported in that function instead?

rgommers:
I think I like `fill_nan(null)`. 

> Is there an API for a more general scalar value replacement? I.E. replace value `a` with value `b` in a Column?

I don't think there is yet. It'd usually be done with indexing like `x[x == 3] = 99`, or with a `where` function/method. `df.where(df == float('nan'), null)` seems unintuitive though and creates a large temporary, so a `fill_nan` method seems preferred.

MarcoGorelli:
sure, let's do fillnan then

just a comment on
```python
df == float('nan')
```
, there'd be no guarantees about this, right? I think people should use `df.isnan()`

rgommers:
> , there'd be no guarantees about this, right? I think people should use `df.isnan()`

[...]

PR #167, opened 2023-05-10
Add a `fill_nan` method to dataframe and column

rgommers:
Addresses half of gh-142 (`fill_null` is more complex, and not included here).

*Note: this is reviewable now, but should be merged after gh-157 which introduces the `null` object).*

### DIFF
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -456,3 +456,17 @@ class Column:
         To get the unique values, you can do ``col.get_rows(col.unique_indices())``.
         """
         ...
+
+    def fill_nan(self, value: float | null, /):
+        """
+        Fill floating point ``nan`` values with the given fill value.
+
+        Parameters
+        ----------
+        value : float or `null`
+            Value used to replace any ``nan`` in the column with. Must be
+            of the Python scalar type matching the dtype of the column (or
+            be `null`).
+
+        """
+        ...
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -457,7 +457,7 @@ class Column:
         """
         ...
 
-    def fill_nan(self, value: float | null, /):
+    def fill_nan(self, value: float | 'null', /):
         """
         Fill floating point ``nan`` values with the given fill value.
 
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -457,7 +457,7 @@ class Column:
         """
         ...
 
-    def fill_nan(self, value: float | 'null', /):
+    def fill_nan(self, value: float | 'null', /) -> Column:
         """
         Fill floating point ``nan`` values with the given fill value.
 
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -707,7 +707,7 @@ class DataFrame:
         """
         ...
 
-    def fill_nan(self, value: float | null, /):
+    def fill_nan(self, value: float | 'null', /):
         """
         Fill ``nan`` values with the given fill value.
 
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -707,7 +707,7 @@ class DataFrame:
         """
         ...
 
-    def fill_nan(self, value: float | 'null', /):
+    def fill_nan(self, value: float | 'null', /) -> DataFrame:
         """
         Fill ``nan`` values with the given fill value.
 
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -706,3 +706,20 @@ class DataFrame:
         In particular, does not check for `np.timedelta64('NaT')`.
         """
         ...
+
+    def fill_nan(self, value: float | null, /):
+        """
+        Fill ``nan`` values with the given fill value.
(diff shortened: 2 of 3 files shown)


jorisvandenbossche (approved):

MarcoGorelli (approved):
looks good (barring docs build error)

rgommers:
It's green now. I had to do the `null` type annotation in string form to make Sphinx happy.

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -706,3 +706,20 @@ def isnan(self) -> DataFrame:
...
+
+        The fill operation will apply to all columns with a floating-point
+        dtype. Other columns remain unchanged.
+
+        Parameters
+        ----------
+        value : float or `null`
+            Value used to replace any ``nan`` in the column with. Must be
+            of the Python scalar type matching the dtype of the column (or
+            be `null`.
```
MarcoGorelli, line 722:
```suggestion
            be `null`).
```

MarcoGorelli (approved):
looks good to me

spec/API_specification/dataframe_api/column_object.py
```diff
@@ -456,3 +456,17 @@ def unique_indices(self, *, skip_nulls: bool = True) -> Column[int]:
         To get the unique values, you can do ``col.get_rows(col.unique_indices())``.
         """
         ...
+
+    def fill_nan(self, value: float | 'null', /):
```
MarcoGorelli, line 460: wait sorry, we need a return type

MarcoGorelli (changes requested):

spec/API_specification/dataframe_api/column_object.py
kkraus14, line 460: A bit unrelated to this PR, but having `null` be typed differently feels like an anti-pattern here. It differentiates between a float scalar (which is implicitly nullable based on our current scalar definition) and a null scalar.

rgommers: > float scalar (which is implicitly nullable based on our current scalar definition

We don't have numpy-style scalars (i.e., instances of a dtype) though? That's why we need a separate `null` object, so that one can construct a column containing nulls with `column_from_sequence([1.5, 2.5, null, 4.5])`.

We could add dtype instances and specify that `null` derives from `float`, but that seems like a huge can of worms for no gain at all. And I think the consensus from what we learned from numpy is that array scalars were a major design mistake.

rgommers: good catch, fixed.

MarcoGorelli: nitpick: we can construct a column with `column_from_sequence([1.5, 2.5, null, 4.5], dtype='float64')` (just pointing this out because it's early days, and I wouldn't want someone to see this and get confused)

MarcoGorelli (approved):
all good, thanks - @kkraus14 any further comments or good to go?

spec/API_specification/dataframe_api/column_object.py
kkraus14: > And I think the consensus from what we learned from numpy is that array scalars were a major design mistake.

Yes, but I think the thing that was agreed as the correct path forward was 0d arrays which we don't have on the DataFrame side. Those 0d arrays are strongly typed and don't have to deal with nulls.

The issue that I see is that someone could do something like:
```
my_int_column = column_from_sequence([1, 2, None, 4], dtype='int32')
max_my_int_column = my_int_column.max(skip_null=False)  # Yields a ducktyped `null` scalar that is int32 typed. Is this `int` type or `null` type from a Typing perspective?

my_float_column = column_from_sequence([1.5, 2.5, max_my_int_column, 4.5], dtype='float64')  # Does this work if the max is `null`?
```

For example, PyArrow handles this by having an explicit `NULL` type (https://arrow.apache.org/docs/python/generated/pyarrow.null.html#pyarrow.null) and presumably has its underlying APIs and compute explicitly handle mixing `NULL` typed scalars / columns with other typed scalars / columns.

Maybe we just need an explicit `NULL` type and then `'null'` here refers to a scalar of type `NULL`?

rgommers: > Maybe we just need an explicit `NULL` type

We do have exactly that already: [docs for `null`](https://data-apis.org/dataframe-api/draft/API_specification/generated/dataframe_api.null.html#dataframe_api.null). The only reason the type annotation is `'null'` rather than `null` is to avoid some circular import and Sphinx weirdness.

kkraus14: That is an object for a `null` scalar as opposed to a `NULL` data type. I.E. allowing a column to be typed `NULL` and extracting a null-valued scalar from that column has type `NULL` versus extracting a null-valued scalar from a float64 column has type `float64`.

It feels counter-intuitive that Columns are type-erased (i.e. just a `Column` class and no `Int32Column`, `Float32Column`, etc.) but the scalars that are contained within Columns are not.

rgommers: Thanks, a new issue sounds good for this. I had not thought before about a need for a null dtype; if there is one we should indeed consider it.

rgommers:
This now has three approvals, so I'll get it in. Thanks all!

spec/API_specification/dataframe_api/column_object.py
jorisvandenbossche: > For example, PyArrow handles this by having an explicit `NULL` type

Small clarification here: while pyarrow indeed has a "null" data type, we also have type-specific null scalars for each data type. And so in your specific example, the `max_my_int_column` would actually be an int32 scalar (with the value of "null"), and not a scalar of the null data type.

PR #171, opened 2023-05-18
Fix __and__ docsting, add __or__

MarcoGorelli:


rgommers (commented):
This really was meant to be `__add__` I suspect, not `__and__`, since it came in with the update to `Column` which matched `DataFrame` methods, and lives right above `__sub__` - and `__add__` was part of the arithmetic methods in gh-94. Also, the docstring says "dataframe" and that should be "column".

Adding `__and__` and `__or__` too seems very reasonable though, I think we can have both `__add__` and `__and__` here.

MarcoGorelli:
good catch, thanks, have updated

rgommers (approved):
LGTM, thanks Marco. The other thing that I thought for a minute should perhaps be different is the error type, however `ValueError` is also used by `any`/`all` when the dtype isn't `bool`. So this should be uncontroversial I think.

rgommers:
I plan to merge this within the next couple of days, unless there is a concern about adding `__and__` and `__or__`.

kkraus14:
Should we clarify whether Kleene logic is used for these or not (to clarify null behavior)?

Also, it looks like for `__and__` we're requiring bools, but for `__or__` it's more generically `Scalar`. Should these be consistent?

MarcoGorelli:
>  but for __or__ it's more generically Scalar.

I think it's also bools there?
```python
def __or__(self, other: Column[bool] | bool) -> Column:
```

it's `__add__` which can more generically take a Scalar

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -410,6 +410,44 @@ def __lt__(self, other: DataFrame | Scalar) -> DataFrame:
...
+            If `self` or `other` is not boolean.
+        """
+
+    def __or__(self, other: DataFrame[bool] | bool) -> DataFrame:
+        """
+        Apply logical 'or' to `other` DataFrame (or scalar) and this DataFrame.
+
+        Parameters
+        ----------
+        other : DataFrame[bool] or Scalar
```
MarcoGorelli, line 438:
```suggestion
        other : DataFrame[bool] or bool
```

rgommers:
> Should we clarify whether Kleene logic is used for these or not (to clarify null behavior)?

That would be good to add indeed. With the answer being that yes, Kleene logic is used I'd think - I assume that's true for all libraries?

Related: we probably should add `__bool__` and make it raise, because truthiness of a column/dataframe is not defined.

jorisvandenbossche:
> I assume that's true for all libraries?

Not for the default (numpy based) data types in pandas. But I would say that's a problem to tackle for pandas / the package implementing this standard on top of pandas. I agree we should specify that it is expected use Kleene logic.

BTW, we should also specify the behaviour of nulls in the comparison methods (`__eq__`, `__lt__`, and friends)

MarcoGorelli:
I've noted that nulls should follow kleene logic - any further requests?

spec/API_specification/dataframe_api/column_object.py
```diff
@@ -165,9 +169,51 @@ def __lt__(self, other: Column | Scalar) -> Column:
         Column
         """
 
-    def __and__(self, other: Column | Scalar) -> Column:
+    def __and__(self, other: Column[bool] | bool) -> Column:
```
kkraus14, line 172: Should we type the output column as `Column[bool]`?

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -410,6 +414,48 @@ def __lt__(self, other: DataFrame | Scalar) -> DataFrame:
         """
         ...
 
+    def __and__(self, other: DataFrame[bool] | bool) -> DataFrame:
```
kkraus14, line 417: Should we type the output as `DataFrame[bool]`?

kkraus14 (approved):

MarcoGorelli:
merging then - thanks for your reviews!

PR #172, opened 2023-05-19
update docs on how to opt-in

### LINKED ISSUE #169: address issue with `__dataframe_namespace__` and non-compliant dataframes
The `__dataframe_namespace__` method is currently modeled on the corresponding method in the array API. However, we're in a slightly different situation here, with a separate developer-facing API. There is a need to convert to/from the `DataFrame` and namespace from a non-standard-compliant user facing dataframe object. We already have the reverse, but this bit is missing - see https://github.com/data-apis/dataframe-api/pull/156#discussion_r1191379962.

Probably the pattern should be:
```python
pdx, df_compliant = df.__dataframe_namespace__()
# use API standard ....
return df_compliant.dataframe
```
when going from public API in a df-using library to the standard-using internals. Inside the internals there should then be only usage of the standard-compliant dataframe ideally.

MarcoGorelli:
closes #169

### DIFF
--- a/protocol/purpose_and_scope.md
+++ b/protocol/purpose_and_scope.md
@@ -55,10 +55,12 @@ def get_compliant_df(df):
     if hasattr(df, '__dataframe_namespace__'):
         # Is already Standard-compliant DataFrame, nothing to do here.
         pass
-    elif hasattr(df, '__dataframe_standard'):
+    elif hasattr(df, '__dataframe_standard__'):
         # Convert to Standard-compliant DataFrame.
         df = df.__dataframe_standard__()
     else:
+        # Here we can raise an exception if we only want to support compliant dataframes,
+        # or convert to our default choice of dataframe if we want to accept (e.g.) dicts
         raise TypeError(
             "Expected Standard-compliant DataFrame, or DataFrame with Standard-compliant implementation"
         )
@@ -67,7 +69,9 @@ def get_compliant_df(df):
 
 def somefunc(df, ...):
     """`df` can be any dataframe conforming to the dataframe API standard"""
+    # Get Standard-compliant DataFrame.
     df = get_compliant_df(df)
+    # Get Standard namespace (optional, only if you need methods from it).
     namespace = df.__dataframe_namespace__()
     # From now on, use `df` methods and `namespace` functions/objects
 ```
--- a/protocol/purpose_and_scope.md
+++ b/protocol/purpose_and_scope.md
@@ -50,24 +50,26 @@ aims to develop in the future. When that full API standard is implemented by
 dataframe libraries, the example above can change to:
 
 ```python
-def get_df_module(df):
+def get_compliant_df(df):
     """Utility function to support programming against a dataframe API"""
     if hasattr(df, '__dataframe_namespace__'):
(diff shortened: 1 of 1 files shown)


protocol/purpose_and_scope.md
```diff
@@ -50,24 +50,26 @@ aims to develop in the future. When that full API standard is implemented by
...
 ```python
-def get_df_module(df):
+def get_compliant_df(df):
     """Utility function to support programming against a dataframe API"""
     if hasattr(df, '__dataframe_namespace__'):
-       # Retrieve the namespace
-       pdx = df.__dataframe_namespace__()
+        # Is already Standard-compliant DataFrame, nothing to do here.
+        pass
+    elif hasattr(df, '__dataframe_standard'):
```
rgommers, line 58: typo: needs `__` after `standard`

protocol/purpose_and_scope.md
```diff
@@ -50,24 +50,26 @@ aims to develop in the future. When that full API standard is implemented by
...
+        # Convert to Standard-compliant DataFrame.
+        df = df.__dataframe_standard__()
     else:
-        # Here we can raise an exception if we only want to support compliant dataframes,
-        # or convert to our default choice of dataframe if we want to accept (e.g.) dicts
```
rgommers, line 60: Can we keep this comment? I thing the change below to raise `TypeError` is good, but the comment is still applicable and informative.

protocol/purpose_and_scope.md
```diff
@@ -50,24 +50,26 @@ aims to develop in the future. When that full API standard is implemented by
...
+        )
+    return df
 
 
 def somefunc(df, ...):
     """`df` can be any dataframe conforming to the dataframe API standard"""
-    pdx, df = get_df_module(df)
-    # From now on, use `df` methods and `pdx` functions/objects
+    df = get_compliant_df(df)
+    namespace = df.__dataframe_namespace__()
```
rgommers, line 71: Now the amount of boilerplate in every function with a dataframe input is 2 lines instead of 1. I'm fairly sure library authors would like to have 1 here, so why not return the namespace from within the utility function?

rgommers (commented):
Thanks Marco. A few questions.

protocol/purpose_and_scope.md
MarcoGorelli: it's only 2 if you need the namespace - for most cases, you probably don't, e.g. for the `z-score` example in https://github.com/MarcoGorelli/impl-dataframe-api/blob/main/blog_post.md, it's not necessary

maybe I'll just comment that it's optional? because returning a tuple from `df, _ = df.__dataframe_standard__` also may feel like boilerplate in all the cases when you don't need the namespace

not really sure here, will think about it

rgommers: Ah yes that is true. Okay, that's a good reason to split it up. +1 for commenting that retrieving the namespace is only needed if one needs one of the handful of items in it.

MarcoGorelli:
any further action needed here?

rgommers (approved):
This LGTM now, thanks @MarcoGorelli

PR #173, opened 2023-05-22
Add a `fill_null` method to dataframe and column

rgommers:
Follow-up to gh-167, which added `fill_nan`, and closes gh-142.

As discussed in gh-142, this does only the scalar case now. I think the main thing that could be done differently is `column_names`. This seemed like the most logical thing to do, keeping `value` to a scalar only. It matches what Vaex does. Pandas instead allows value to be: `scalar, dict, Series, or DataFrame`. Polars doesn't seem to have any way of dealing with non-uniform dtypes of columns.

### DIFF
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -641,3 +641,16 @@ class Column:
 
         """
         ...
+
+    def fill_null(self, value: Scalar, /) -> Column:
+        """
+        Fill null values with the given fill value.
+
+        Parameters
+        ----------
+        value : Scalar
+            Value used to replace any ``null`` values in the column with.
+            Must be of the Python scalar type matching the dtype of the column.
+
+        """
+        ...
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -774,3 +774,36 @@ class DataFrame:
 
         """
         ...
+
+    def fill_null(
+        self, value: Scalar, /, *, column_names : list[str] | None = None
+    ) -> DataFrame:
+        """
+        Fill null values with the given fill value.
+
+        This method can only be used if all columns that are to be filled are
+        of the same dtype kind (e.g., all floating-point, all integer, all
+        string or all datetime dtypes). If that is not the case, it is not
+        possible to use a single Python scalar type that matches the dtype of
+        all columns to which ``fill_null`` is being applied, and hence an
+        exception will be raised.
+
+        Parameters
+        ----------
+        value : Scalar
+            Value used to replace any ``null`` values in the dataframe with.
+            Must be of the Python scalar type matching the dtype(s) of the dataframe.
+        column_names : list[str] | None
+            A list of column names for which to replace nulls with the given
+            scalar value.
+
+        Raises
+        ------
+        TypeError
+            If the columns of the dataframe are not all of the same kind.
+        KeyError
+            If ``column_names`` contains a column name that is not present in
+            the dataframe.
+
+        """
+        ...
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -782,11 +782,10 @@ class DataFrame:
         Fill null values with the given fill value.
 
         This method can only be used if all columns that are to be filled are
-        of the same dtype kind (e.g., all floating-point, all integer, all
-        string or all datetime dtypes). If that is not the case, it is not
-        possible to use a single Python scalar type that matches the dtype of
-        all columns to which ``fill_null`` is being applied, and hence an
-        exception will be raised.
+        of the same dtype (e.g., all of ``Float64`` or all of string dtype).
+        If that is not the case, it is not possible to use a single Python
+        scalar type that matches the dtype of all columns to which
+        ``fill_null`` is being applied, and hence an exception will be raised.
 
         Parameters
         ----------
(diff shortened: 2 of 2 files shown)


spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -723,3 +723,36 @@ def fill_nan(self, value: float | 'null', /) -> DataFrame:
...
+        exception will be raised.
+
+        Parameters
+        ----------
+        value : Scalar
+            Value used to replace any ``null`` values in the dataframe with.
+            Must be of the Python scalar type matching the dtype(s) of the dataframe.
+        column_names : list[str] | None
+            A list of column names for which to replace nulls with the given
+            scalar value.
```
MarcoGorelli, line 747: if `None`, does it apply to all columns?

MarcoGorelli (commented):
small comment, but generally in favour

spec/API_specification/dataframe_api/dataframe_object.py
rgommers: yes it does - will clarify

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -723,3 +723,36 @@ def fill_nan(self, value: float | 'null', /) -> DataFrame:
...
+
+    def fill_null(
+        self, value: Scalar, /, *, column_names : list[str] | None = None
+    ) -> DataFrame:
+        """
+        Fill null values with the given fill value.
+
+        This method can only be used if all columns that are to be filled are
+        of the same dtype kind (e.g., all floating-point, all integer, all
+        string or all datetime dtypes). If that is not the case, it is not
```
kkraus14, line 735: Should we clarify whether we expect them to all be the same bit-width? I.E. how this current reads is that a mix of `int32` and `int64` is potentially allowed.

rgommers: I did intend to allow that, because the input is a Python `int` or `float` either way. I don't feel strongly about it though; I think it works as is, but if we want to be more strict then we can do that (loosening later is always possible).

kkraus14: > I did intend to allow that, because the input is a Python `int` or `float` either way.

It's a ducktyped Python `int` or `float` that is potentially strongly typed under the hood and the casting could be potentially lossy in the case of `float64` --> `float32` or datetime resolution downgrade `ns` --> `us`.

I'd prefer to be more strict now and figure out the path towards loosening later.

MarcoGorelli: can we move forwards with this? is the solution just to replace "dtype kind" with "dtype"?

rgommers: > is the solution just to replace "dtype kind" with "dtype"?

Yes I think so - done now.

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -723,3 +723,36 @@ def fill_nan(self, value: float | 'null', /) -> DataFrame:
...
+        possible to use a single Python scalar type that matches the dtype of
+        all columns to which ``fill_null`` is being applied, and hence an
+        exception will be raised.
+
+        Parameters
+        ----------
+        value : Scalar
+            Value used to replace any ``null`` values in the dataframe with.
+            Must be of the Python scalar type matching the dtype(s) of the dataframe.
+        column_names : list[str] | None
```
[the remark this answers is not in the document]
rgommers: We currently don't use either of `column_names` or `subset`. I'm happy either way, I think the former is a bit more explicit about what it refers to while the latter expresses better what is done with the provided names. 

Anyone else have a preference?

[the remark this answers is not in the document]
MarcoGorelli: this seems fine

MarcoGorelli (approved):

spec/API_specification/dataframe_api/column_object.py
```diff
@@ -641,3 +641,16 @@ def fill_nan(self, value: float | 'null', /) -> Column:
 
         """
         ...
+
+    def fill_null(self, value: Scalar, /) -> Column:
```
kkraus14, line 645: Not the PR to discuss it but it would be really nice if for typing purposes we could type this as:
```
    def fill_null(self[T], value: Scalar[T], /) -> Column[T]:
```

kkraus14 (approved):

spec/API_specification/dataframe_api/column_object.py
rgommers: gh-187 makes a big step in this direction. That said, the generics don't quite work here I think, since the `T` in `Scalar[T]` is a Python scalar while in `Column[T]` it's a dtype. We happened to have a good discussion about gh-187 yesterday, and striking a balance between making the spec fully type-checkable and keeping it readable. I suspect this kind of case requires overloads, which may be useful at some point but can't be rendered as html - so it divorces static typing a bit from spec writing.

MarcoGorelli: I think there should be a way to make `Scalar[T]` work, and every library can define what their scalar type is corresponding to each of the standard's types. But that's for another issue, let's pick it up again after https://github.com/data-apis/dataframe-api/pull/187 is in (I'll update it today)

rgommers: > and every library can define what their scalar type is corresponding to each of the standard's types

Quick comment: they are, or at least include, Python builtin types, which one has no control over. So it'd be `int) -> Column[Int64]` or some such thing.

> But that's for another issue, let's pick it up again after #187 is in (I'll update it today)

Sounds good, thanks.

PR #175, opened 2023-05-24
add `Column.__dataframe_namespace__`

MarcoGorelli:
Pretty minor, hope this is uncontroversial

use-case: you have a function which accepts a Column as input, and you need to access some method from the top-level namespace

### DIFF
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import NoReturn, Sequence, TYPE_CHECKING
+from typing import Any, NoReturn, Sequence, TYPE_CHECKING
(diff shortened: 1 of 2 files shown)


rgommers (changes requested):
I understand the use case, but https://data-apis.org/dataframe-api/draft/purpose_and_scope.html#checking-a-dataframe-object-for-compliance strongly implies that this method should only be implemented by standard-compliant dataframe objects, and that it's the way to check whether an object is such a dataframe.

I think it should probably be `__column_namespace__` instead, as the one way to check if this is a standard-compliant `Column` object. And it additionally can guarantee that it will return the same namespace as for `DataFrame.__dataframe_namespace__`.

rgommers (approved):
LGTM now, in it goes. Thanks Marco!

PR #178, opened 2023-06-15
use snake case

MarcoGorelli:
resurrecting #111 as this has been annoying me slightly

it's true that python numpy and pandas are inconsistent here, but we have chance to do things differently and better in this standard - I suggest we just use snake case consistently

### DIFF
--- a/spec/API_specification/dataframe_api/__init__.py
+++ b/spec/API_specification/dataframe_api/__init__.py
@@ -17,7 +17,7 @@ __all__ = [
     "column_from_sequence",
     "concat",
     "dataframe_from_dict",
-    "isnull",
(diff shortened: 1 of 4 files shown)


jorisvandenbossche:
+1, especially given that we also use `skip_nulls`

rgommers (approved):
Thanks Marco. I'm happy either way here; it's a stylistic thing and I've been trying to retreat from having strong opinions on those in all my projects:) So I'll just add my approval review here. Will merge within the next day or two unless there's more discussion.

MarcoGorelli:
thanks for your reviews

merging then to keep things moving, can revert if there's objections

PR #179, opened 2023-06-15
Add some dtypes

### LINKED ISSUE #158: How to type return value of `Column` reductions
This is related to https://github.com/data-apis/dataframe-api/issues/155, but I'm opening a new issue to make sure it's not forgotten

Currently column reductions are typed to return `type`. But if we also use `type` to type arguments, such as `Column.from_sequence`, then they should probably be distinguished.

For example, I could pass `'float64'` to `Column.from_sequence`, but the return value of `column.mean()` wouldn't be `'float64'`, it would probably just be a Python 'float'.

I don't really know yet what the solution here is, will think about it, but I think we can do better

MarcoGorelli:
I'm adding a `DType` class, as well as some classes for common dtypes

I'm also correcting the return type of some methods - for example, `min` doesn't return a dtype, it returns a scalar

closes #158


spec/API_specification/dataframe_api/__init__.py
```diff
@@ -19,6 +17,17 @@
...
+    "Int64",
+    "Int32",
+    "Int16",
+    "Int8",
+    "UInt64",
+    "UInt32",
+    "UInt16",
+    "UInt8",
+    "Float64",
+    "Float32",
```
kkraus14, line 30: Is `Bool` missing from here?

spec/API_specification/dataframe_api/__init__.py
```diff
@@ -129,3 +138,43 @@ def isnull(value: object, /) -> bool:
...
+    """Unsigned integer type with 8 bits of precision."""
+
+class Float64(DType):
+    """Floating point type with 64 bits of precision."""
+
+class Float32(DType):
+    """Floating point type with 32 bits of precision."""
+
+class Bool(DType):
+    """Boolean type."""
```
kkraus14, line 180: Should we have a `Bool8` vs `Bool1`? This is a place where some libraries differ in their representations.

MarcoGorelli: sure

jorisvandenbossche: The dtypes aren't meant to be "logical"? In which case bool1 vs bool8 doesn't matter, it's just bool and the implementation can do what they want to give the behaviour of a boolean column

MarcoGorelli: hmm yeah, unless all libraries have both, then I don't think anyone would use `Bool8` or `Bool1` when using the standard?

kkraus14 (approved):

MarcoGorelli:
thanks for your review!

PR #182, opened 2023-06-20
Note __pow__ return type

### LINKED ISSUE #174: Return type of pow
Seems like there's some inconsistencies:

```python
In [6]: pd.Series([1,2,3])**2
Out[6]:
0    1
1    4
2    9
dtype: int64

In [7]: pd.Series([1,2,3])**-1
ValueError: Integers to negative integer powers are not allowed.

In [8]: pl.Series([1,2,3]).pow(2)
Out[8]:
shape: (3,)
Series: '' [f64]
[
        1.0
        4.0
        9.0
]

In [9]: pl.Series([1,2,3]).pow(-1)
Out[9]:
shape: (3,)
Series: '' [f64]
[
        1.0
        0.5
        0.333333
]
```

polars always returns floats, whereas pandas returns either integers or floats, and may error based on the value of the exponent

What do we want to do here?

MarcoGorelli:
closes #174

### DIFF
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -295,6 +295,9 @@ class Column:
         """
         Raise this column to the power of `other`.
 
+        Integer to the power of non-negative integer is integer.
+        Float to the power of integer or float is float.
+
         Parameters
         ----------
         other : Column or Scalar
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -296,6 +296,7 @@ class Column:
         Raise this column to the power of `other`.
 
         Integer dtype to the power of non-negative integer dtype is integer dtype.
+        Integer dtype to the power of float dtype is float dtype.
         Float dtype to the power of integer dtype or float dtype is float dtype.
 
         Parameters
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -295,8 +295,8 @@ class Column:
         """
         Raise this column to the power of `other`.
 
-        Integer to the power of non-negative integer is integer.
-        Float to the power of integer or float is float.
+        Integer dtype to the power of non-negative integer dtype is integer dtype.
+        Float dtype to the power of integer dtype or float dtype is float dtype.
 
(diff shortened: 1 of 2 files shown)


rgommers (commented):
"integer" and "float" are perhaps not precise enough (those are Python scalar types, not dataframe dtypes). How about saying something about the input dtype being preserved in all cases if the right-hand operand is integer?

MarcoGorelli:
is it always preserved though? upcasting could happen, such as int8 -> int16. maybe that doesn't need standardising on, but noting that integer dtype stays integer dtype does

spec/API_specification/dataframe_api/column_object.py
```diff
@@ -295,6 +295,9 @@ def __pow__(self, other: Column | Scalar) -> Column:
         """
         Raise this column to the power of `other`.
 
+        Integer dtype to the power of non-negative integer dtype is integer dtype.
```
kkraus14, line 298: What is the behavior of integer dtype to the power of float dtype?

kkraus14, line 298:
```suggestion
        Integer dtype to the power of non-negative integer dtype is integer dtype.
        Integer dtype to the power dtype of float is float dtype.
```

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -528,6 +528,10 @@ def __pow__(self, other: DataFrame | Scalar) -> DataFrame:
         """
         Raise this dataframe to the power of `other`.
 
+        Integer to the power of non-negative integer is integer.
+        Integer to the power of float is float.
+        Float to the power of integer or float is float.
```
kkraus14, line 533:
```suggestion
        Integer dtype to the power of non-negative integer dtype is integer dtype.
        Integer dtype to the power of float dtype is float dtype.
        Float dtype to the power of integer dtype or float dtype is float dtype.
```

kkraus14 (approved):

spec/API_specification/dataframe_api/column_object.py
MarcoGorelli: sure, thanks

rgommers (approved):

MarcoGorelli:
merging, thanks all

PR #183, opened 2023-06-20
Add a `correction` keyword to the `std` methods

### LINKED ISSUE #170: std / var : ddof?
Doesn't look like the current std and var functions have a ddof argument - I presume they should?

MarcoGorelli:
closes #170

### DIFF
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -479,8 +479,18 @@ class Column:
         Parameters
         ----------
         correction
-            Correction to apply to the result. For example, 0 for sample standard
-            deviation and 1 for population standard deviation.
+            Degrees of freedom adjustment. Setting this parameter to a value other
+            than ``0`` has the effect of adjusting the divisor during the
+            calculation of the standard deviation according to ``N-correction``,
+            where ``N`` corresponds to the total number of elements over which
+            the standard deviation is computed. When computing the standard
+            deviation of a population, setting this parameter to ``0`` is the
+            standard choice (i.e., the provided column contains data
+            constituting an entire population). When computing the corrected
+            sample standard deviation, setting this parameter to ``1`` is the
+            standard choice (i.e., the provided column contains data sampled
+            from a larger population; this is commonly referred to as Bessel's
+            correction). Fractional (float) values are allowed. Default: ``1``.
         skip_nulls
             Whether to skip null values.
         """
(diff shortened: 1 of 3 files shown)


spec/API_specification/dataframe_api/column_object.py
```diff
@@ -410,20 +410,36 @@ def mean(self, *, skip_nulls: bool = True) -> DType:
         dtypes.
         """
 
-    def std(self, *, skip_nulls: bool = True) -> DType:
+    def std(self, *, correction: int | float = 1, skip_nulls: bool = True) -> DType:
```
rgommers, line 413: This should use `Optional` or `| None` it looks like. Same for `var` further down.

rgommers (commented):
Thanks Marco. LGTM modulo my minor type annotation comment.

spec/API_specification/dataframe_api/column_object.py
MarcoGorelli: why should it allow `None`, instead of just putting a default value?

spec/API_specification/dataframe_api/column_object.py
```diff
@@ -469,20 +469,36 @@ def mean(self, *, skip_nulls: bool = True) -> Scalar:
...
         Reduction returns a scalar. Must be supported for numerical and
         datetime data types. Returns a float for numerical data types, and
         datetime (with the appropriate timedelta format string) for datetime
         dtypes.
+
+        Parameters
+        ----------
+        correction
+            Correction to apply to the result. 0 for sample standard deviation
+            and 1 for population standard deviation.
```
kkraus14, line 483: If we're only supporting 0 or 1 can this just be an `int` for now? We could change to allow `float`s later if desired?

kgryte: I think the description is misleading here. The `0` and `1` references are, tmk, to help users identify common correction values. See the [Array API specification](https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.std.html) for a more extensive description.

kgryte: Fine with me, but up to Marco. :)

MarcoGorelli: I've reworded to make clear that `0` and `1` are just examples

rgommers: That rewording was for only 1 of 4 docstrings. I now edited all of them, and for one adopted the more extensive description from the array API standard.

rgommers:
Also a note on the default: that is `1` (which matches Pandas) rather than `0` (which matches the array API standard and numpy). This seems fine to me, just pointing it out for future reference.

spec/API_specification/dataframe_api/column_object.py
rgommers: @MarcoGorelli if you're happy with that change, I think this PR is good to go.

MarcoGorelli:
much appreciated, nice improvement - thanks!

PR #187, opened 2023-06-24
Make the spec implementation in this repo type checkable

MarcoGorelli:
Currently, the repo can't be type-checked. In particular, it can't be used to verify whether implementations conform to it

Let's change that! 🚀 

Summary of changes:
- adding a generic type for `DataFrame`, `Series`, and `Scalar`, so we can use that in type annotations. If you see a `Column` method like `def foo(self) -> Column[T]`, it means it returns a column of the same dtype

This helped find quite a few missing bits in https://github.com/MarcoGorelli/impl-dataframe-api


rgommers:
> renaming the files `.py` -> `.pyi` (so their bodies don't need filling in)

I'm not sure this can work with the doc build. Unless Sphinx gained some capability to read docstrings from `.pyi` files? I'm not aware of docstrings in `.pyi` files being a thing at all.

MarcoGorelli:
yup, that's right, thanks - I've gone for using `mypy.ini` to ignore the `empty-body` code instead

MarcoGorelli:
I've updated to make `DataFrame` no longer generic, this simplifies things a bit

Sometimes I've used `Column[Any]` when perhaps something more precise would have been used, but for now I'd suggest we start with this so we at least have something type-checkable which we can use to test implementations

Any objections to getting this in?

rgommers (commented):
Thanks @MarcoGorelli. The `Column[DtypeT]` and `DataFrame` changes look quite good - nice to see that things are still readable _and_ they now type-check.

The one thing that raises questions for me is `class Scalar(Generic[DTypeT]):` and using things like `Scalar[Bool]`. That does not seem right, or at a minimum confusing, because there are no scalar instances in the spec - where we used `Scalar` till now we really mean instances of Python `bool`, `int` or `float`, with the caveat that they may be ducktyped. The way you have it now seems to hint at "dataframe scalars" a la numpy's "array scalars".

rgommers:
> we really mean instances of Python `bool`, `int` or `float`, with the caveat that they may be ducktyped.

See https://data-apis.org/dataframe-api/draft/design_topics/python_builtin_types.html

MarcoGorelli:
yeah arguably that part still needs sorting out a bit

if the spec specified `int`, and some implementation returned a ducktyped `int`, would it still type-check?

rgommers:
> if the spec specified `int`, and some implementation returned a ducktyped `int`, would it still type-check?

Hard to know without trying. And if it wouldn't, whether the right approach to fix it is in the spec or in the duck typing object by declaring it compatible with the builtin `int` somehow.

For your current purposes - testing the Pandas and Polars implementations - it should not matter, right?

MarcoGorelli:
practically speaking, I don't think it works to make `Scalar` generic, we may need:
- `namespace.IntScalar`
- ...

rgommers:
I think that should be:
```python
def get_value(self, row_number: Scalar[IntDType]) -> int:
```
while for something like this:
```python
def some_method(self, param: Scalar[DType]) -> Scalar[DType]:
```
that's where I'd expect a clash between typing needs (which would need to use `@overload` to support `int`/`float`/etc return types) and a human-readable spec.

> we may need `namespace.IntScalar`

That would be potentially confusing I think, and is the kind of issue where static typing tends to be counter-productive. There is no such thing as `IntScalar` in the API, it's simply the builtin `int` (and if there's a duck typing thing, the library may choose to keep that private). Trying to statically type things imho conflicts with writing a clear spec at that point.

The workaround depends on how well Sphinx works now with type aliases. See [`autodoc_type_aliases`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_type_aliases) & co. It's not yet clear what the actual type annotation should be, but using support for type aliases it may be possible to have the spec read `int` and the annotation such that both `int` and duck-typed ints work.

MarcoGorelli:
> def get_value(self, row_number: Scalar[IntDType]) -> int:

is there a typo here? I thought you wanted to get rid of the generic `Scalar`, and also because the return type isn't `int`, is of the dtype of the column

e.g.:
```python
>>> df = pl.DataFrame({'a': [False, True]})
>>> convert_to_standard_compliant_dataframe(df)
<polars_standard.polars_standard.PolarsDataFrame object at 0x7f0c47abad90>
>>> convert_to_standard_compliant_dataframe(df).get_column_by_name('a').get_value(0)
False
```

rgommers:
> is there a typo here? I thought you wanted to get rid of the generic `Scalar`, and also because the return type isn't `int`, is of the dtype of the column

argh yes, I was focusing on the return type only. And I got it wrong because type annotations are obscure:) So let me try again:
```python
def get_value(self, row_number: int) -> Scalar:
```
I don't see a way to remove `-> Scalar` there in a way that doesn't make the html docs more difficult to read. The current docs seem clear to me: https://data-apis.org/dataframe-api/draft/API_specification/column_object.html#dataframe_api.Column.get_value. 

My suggestion is to leave that part as-(unchanged from `main`) in this PR, and in a follow-up one figure out how to use the `autodoc` support for type aliases to improve type-checkability without changing the rendered html. (or do it in this PR if you prefer)

MarcoGorelli:
Sure, I've reverted making `Scalar` generic for now, let's revisit separately

spec/API_specification/index.rst
```diff
@@ -28,6 +28,9 @@ of objects and functions in the top-level namespace. The latter are:
    Float64
    Float32
    Bool
+   DTypeT
+   IntDType
+   FloatDType
```
rgommers, line 33: I think these 3 items should not be added, since they're not part of the Python API but rather only `TypeVar`'s.

spec/API_specification/dataframe_api/__init__.py
```diff
@@ -29,6 +31,9 @@
     "Float64",
     "Float32",
     "Bool",
+    "DTypeT",
+    "FloatDType",
+    "IntDType",
```
rgommers, line 36: These things aren't public API, so shouldn't be added here.

rgommers (commented):
Two more comments about not including typevars and classes that are introduced only for the purpose of static typing as public Python API. I think/hope you can just remove them without Sphinx becoming unhappy. Then I think this is good to go.

spec/API_specification/index.rst
rgommers: > isn't it useful at runtime to tell whether something is any integer dtype?

It may seem like it, but deciding that we want a dtype class hierarchy is a potentially very large can of worms. Python's "numeric tower" is basically considered a design mistake at this point (Guido has stated as much, and there's long blog posts you can find about it) and the array API does not have a class hierarchy on purpose, and has an `isdtype` function to address the `isinstance` check use case.

At a minimum, doing this requires a new issue and design discussion, we can't just put merge that in right now to make type checking easier.

MarcoGorelli: Sure, we can discuss in the next call

I presume we can't export `Scalar` either then? I'll mark as `Any` for now then, otherwise it won't work for type-checking implementations

MarcoGorelli:
Thanks, have updated

I've marked a few types as `Any` (hopefully temporarily) until #188 and #189 are resolved, but this at least makes it so that implementations can use it

rgommers:
Did you try making `Scalar` a [type alias](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#type-aliases) of `Any`? E.g., Mypy seems perfectly happy with this:
```diff
diff --git a/spec/API_specification/dataframe_api/column_object.py b/spec/API_specification/dataframe_api/column_object.py
index 3880406..145ac8a 100644
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -11,6 +11,11 @@ if TYPE_CHECKING:
 __all__ = ['Column']
 
 
+# Type alias: Mypy needs Any, but for readability we need to make clear this
+# is a Python scalar (i.e., an instance of `bool`, `int`, `float`, `str`, etc.)
+Scalar = Any
+
+
 class Column(Generic[DType]):
     """
     Column object
@@ -138,7 +143,7 @@ class Column(Generic[DType]):
         """
         ...
 
-    def __eq__(self: Column[DType], other: Column[DType] | Any) -> Column[Bool]:  # type: ignore[override]
+    def __eq__(self: Column[DType], other: Column[DType] | Scalar) -> Column[Bool]:  # type: ignore[override]
         """
         Compare for equality.
```
Typing-wise it's fully equivalent, and it avoids throwing away the intent. It'd also make the diff of this PR quite a bit smaller.

rgommers:
Besides that issue, this looks pretty good to me now.

spec/API_specification/dataframe_api/__init__.py
```diff
@@ -59,7 +60,7 @@ def concat(dataframes: Sequence[DataFrame]) -> DataFrame:
     """
     ...
 
-def column_from_sequence(sequence: Sequence[object], *, dtype: DType) -> Column:
+def column_from_sequence(sequence: Sequence[Any], *, dtype: Any) -> Column[Any]:
```
jorisvandenbossche, line 63: `dtype: DType` seems more specific that it actually needs to be a dtype instance, and not eg a string repr?

spec/API_specification/dataframe_api/column_object.py
```diff
@@ -137,7 +138,7 @@ def sorted_indices(
         """
         ...
 
-    def __eq__(self, other: Column | Scalar) -> Column:
+    def __eq__(self: Column[DType], other: Column[DType] | Any) -> Column[Bool]:  # type: ignore[override]
```
jorisvandenbossche, line 141: Not familiar with those details of typing, but using `Column[DType]` twice, does that indicate both columns need to have the same dtype? Or they can both still have their own dtype?

spec/API_specification/dataframe_api/__init__.py
rgommers: I noticed that too, but I think it can't be addressed with a type alias like I suggested for `Scalar`, because `DType` is already a `TypeVar`.

spec/API_specification/dataframe_api/column_object.py
MarcoGorelli: yeah I think we'd be expecting them to have the same dtype in order to compare them?

spec/API_specification/dataframe_api/__init__.py
MarcoGorelli: let's address https://github.com/data-apis/dataframe-api/issues/189 first, then we can figure out how to type this

for now I've marked as `Any` to not block the benefits which this PR brings

spec/API_specification/dataframe_api/column_object.py
jorisvandenbossche: That would not be my expectation, eg numpy/pandas (and I suppose the array API as well?) supports comparing any numeric dtype with any other numeric dtype.

rgommers (approved):
This looks quite good now, CI is green and it's helpful for prototyping in https://github.com/data-apis/dataframe-api-compat/ - so let's get this in. Thanks @MarcoGorelli!

PR #192, opened 2023-06-29
Clarify `__dataframe_standard__`, introduce `__column_standard__`

### LINKED ISSUE #190: How to convert Series to compliant Column?
Say I have a function which takes a pandas Series:
```python
def calculate_std_of_series(ser: pd.Series):
    return ser.std(ddof=1)
```
I'd like to convert this to support any standard-compliant library. How can I do that?

I'd like to be able to do something like
```python
def calculate_std_of_series(ser: Any):
    ser_compliant = ser.__series_standard__()
    return ser_compliant.std(correction=1)
```

Currently, I can't do that, because we don't have `__series_standard__`.

Are we OK to require the libraries implementing the Standard have a `__series_standard__` method in whichever object they use to back their `Column` objects?

---

Context of why this is necessary: in plotly, it's possible to pass a Series to `'x'` / `'y'` / etc:
```python
df = pd.DataFrame({'x': [1, 2, 3], 'y': [1, 3, 2]})
px.line(df, x='x', y=pd.Series([1,2,3]))
```

I expect there'll initially be pushback (just something I've come to expect 😄 ) - if you do disagree that we should do have `__series_standard__`, could you please suggest an alternative for writing library-agnostic functions which operate on Series?
And if the answer is that we don't want to support that, then could you please suggest how plotly could support
```python
px.line(df, x='x', y=pd.Series([1,2,3]))
```
in a library-agnostic manner?

MarcoGorelli:
closes #190

### DIFF
--- a/protocol/purpose_and_scope.md
+++ b/protocol/purpose_and_scope.md
@@ -76,14 +76,6 @@ def somefunc(df, ...):
     # From now on, use `df` methods and `namespace` functions/objects
 ```
 
-Libraries which implement the Standard in a separate namespace
-are required to provide the following methods:
-- ``__dataframe_standard__``: used for converting a non-compliant dataframe to a compliant one;
-- ``__column_standard__``: used for converting a non-compliant column to a compliant one.
-
-For example, pandas would have ``pandas.DataFrame.__dataframe_standard__`` and
-``pandas.Series.__column_standard__``.
-
 ### History
 
 Dataframe libraries in several programming language exist, such as
--- a/protocol/purpose_and_scope.md
+++ b/protocol/purpose_and_scope.md
@@ -76,6 +76,13 @@ def somefunc(df, ...):
     # From now on, use `df` methods and `namespace` functions/objects
 ```
 
+Libraries which implement the Standard in a separate namespace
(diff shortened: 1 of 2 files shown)


protocol/purpose_and_scope.md
```diff
@@ -76,6 +76,13 @@ def somefunc(df, ...):
     # From now on, use `df` methods and `namespace` functions/objects
 ```
 
+Libraries which implement the Standard in a separate namespace
+are required to provide the following methods:
+- ``__dataframe_standard__``: used for converting a non-compliant dataframe to a compliant one;
+- ``__column_standard__``: used for converting a non-compliant column to a compliant one.
+
+For example, pandas would have ``pandas.DataFrame.__dataframe_standard__`` and
+``pandas.Series.__column_standard__``.
```
jorisvandenbossche, line 85: This should probably be moved to the spec part of the repo?

MarcoGorelli: good point, thanks

rgommers (approved):
LGTM, thanks for helping clear this up @MarcoGorelli and @jorisvandenbossche

PR #193, opened 2023-07-03
Add some cumulative Column operations

MarcoGorelli:
I noticed plotly uses some cumulative functions (`Series.cumsum`) - OK to start adding some simple and common cumulative operations to `Column`?

### DIFF
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -525,26 +525,26 @@ class Column:
             Whether to skip null values.
         """
 
-    def cummax(self) -> Column:
+    def cumulative_max(self) -> Column:
         """
         Reduction returns a Column. Any data type that supports comparisons
         must be supported. The returned value has the same dtype as the column.
         """
 
-    def cummin(self) -> Column:
+    def cumulative_min(self) -> Column:
         """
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
If we add them, we might want to use the full name (cumulative_...), xref https://github.com/data-apis/array-api/issues/597#issuecomment-1431152772

kgryte:
These APIs are likely to be added to the 2023 revision of the array API standard. May be preferable to either wait until the naming convention is sorted there, or help participate in that discussion. Apart from `cumulative_sum`, another alternative is `cusum`.

MarcoGorelli:
have updated, thanks

let's go with `cumulative_`, and we can always update later again if necessary?

rgommers:
> let's go with `cumulative_`, and we can always update later again if necessary?

That sounds fine to me - keeps the ball rolling.

rgommers (approved):
This seems fine to me, thanks Marco. Assuming folks want the min/max too; they seem commonly implemented in dataframe libraries at least. 

And checking: typing is going to need `Column[Any]` perhaps, given the other typing PR?

MarcoGorelli:
for sum and prod, yes

for `cumulative_min` and `cumulative_max`, I think we could be more precise: `Column[DType]` (it should be preserved for these two operations, right?)

PR #197, opened 2023-07-12
Add some more dtypes: Date, Datetime, Duration, String

MarcoGorelli:
I suggest we leave categorical for a separate PR (will probably require discussion)

---

EDIT: marking as draft for now, let's leave this out from the first version

### DIFF
--- a/spec/API_specification/dataframe_api/__init__.py
+++ b/spec/API_specification/dataframe_api/__init__.py
@@ -30,7 +30,6 @@ __all__ = [
     "Float64",
     "Float32",
     "Bool",
-    "Date",
     "Datetime",
     "Duration",
     "String",
@@ -182,9 +181,6 @@ class Float32:
 class Bool:
     """Boolean type."""
 
-class Date:
-    """Date type."""
-
 class Datetime:
     """
     Datetime type.
@@ -192,7 +188,8 @@ class Datetime:
     Attributes
     ----------
     time_unit : Literal['ms', 'us', 'ns']
-        Precision of the datetime type.
+        Precision of the datetime type. There is no guarantee that the full
+        range of dates available for the specified precision is supported.
     time_zone : str | None
         Time zone of the datetime type. Only IANA time zones are supported.
         `None` indicates time-zone-naive data.
--- a/spec/API_specification/dataframe_api/__init__.py
+++ b/spec/API_specification/dataframe_api/__init__.py
@@ -180,7 +180,7 @@ class Float32:
     """Floating point type with 32 bits of precision."""
 
 class Bool:
-    """Boolean type with 8 bits of precision."""
+    """Boolean type."""
 
 class Date:
     """Date type."""
@@ -201,7 +201,7 @@ class Datetime:
     time_zone: str | None  # Only IANA time zones are supported
 
 class Duration:
-    """Duration type with 64 bits of precision."""
+    """Duration type."""
     time_unit: Literal['ms', 'us', 'ns']
 
 class String:
--- a/spec/API_specification/dataframe_api/__init__.py
+++ b/spec/API_specification/dataframe_api/__init__.py
@@ -3,7 +3,7 @@ Function stubs and API documentation for the DataFrame API standard.
 """
 from __future__ import annotations
 
-from typing import Mapping, Sequence, Any
+from typing import Mapping, Sequence, Any, Literal
 
 from .column_object import *
 from .dataframe_object import DataFrame
@@ -177,3 +177,25 @@ class Float32:
 
 class Bool:
     """Boolean type with 8 bits of precision."""
+
+class Datetime:
+    """
+    Datetime type with 64 bits of precision.
+
+    Attributes
+    ----------
+    time_unit : Literal['ms', 'us', 'ns']
+        Precision of the datetime type.
+    time_zone : str | None
+        Time zone of the datetime type. Only IANA time zones are supported.
+        `None` indicates time-zone-naive data.
+    """
+    time_unit: Literal['ms', 'us', 'ns']
+    time_zone: str | None  # Only IANA time zones are supported
+
+class Duration:
+    """Duration type with 64 bits of precision."""
+    time_unit: Literal['ms', 'us', 'ns']
+
+class String:
+    """String type."""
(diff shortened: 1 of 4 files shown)


rgommers (commented):
These names and attributes look good to me - readable/understandable. My main question right now: did you choose the available units as the minimum common set that all relevant libraries support today? Arrow has some other ones, like seconds for `Datetime` and days/milliseconds(?) for `Date`, and months for `Duration: https://arrow.apache.org/docs/format/CDataInterface.html#data-type-description-format-strings

MarcoGorelli:
Yes, exactly - technically even `Date` would be a stretch at the moment, as it's not in pandas, but I think we can work around that (by just always validating that anything sub-hourly is zero)

's' resolution in polars would be quite tricky as `chrono` is currently limited to dates in the -280,000, 280,000 range https://docs.rs/chrono/latest/chrono/#limitations

We could just add `'s'` to the standard and note that the full range of dates for that unit may not necessarily be reached for every implementation?

spec/API_specification/dataframe_api/__init__.py
```diff
@@ -177,3 +181,28 @@ class Float32:
 
 class Bool:
     """Boolean type with 8 bits of precision."""
+
+class Date:
+    """Date type."""
+
+class Datetime:
+    """
+    Datetime type.
```
kkraus14, line 190: Should we specify how this is represented?

spec/API_specification/dataframe_api/__init__.py
```diff
@@ -177,3 +181,28 @@ class Float32:
...
+        Precision of the datetime type.
+    time_zone : str | None
+        Time zone of the datetime type. Only IANA time zones are supported.
+        `None` indicates time-zone-naive data.
+    """
+    time_unit: Literal['ms', 'us', 'ns']
+    time_zone: str | None  # Only IANA time zones are supported
+
+class Duration:
+    """Duration type with 64 bits of precision."""
```
kkraus14, line 204: Should we clarify that this is a 64 bit integer? I've seen other implementations floating around that do something like use a struct of multiple integers to represent the different unit granularities separately

spec/API_specification/dataframe_api/__init__.py
```diff
@@ -177,3 +181,28 @@ class Float32:
 
 class Bool:
     """Boolean type with 8 bits of precision."""
+
+class Date:
+    """Date type."""
```
kkraus14, line 186: Should we specify how this is represented?

spec/API_specification/dataframe_api/__init__.py
```diff
@@ -177,3 +181,28 @@ class Float32:
 
 class Bool:
     """Boolean type with 8 bits of precision."""
```
kkraus14, line 183: I think we previously punted on defining the bitwidth of this since it's for the compute API as opposed to interchange protocol? Should we unspecify the "8 bits of precision" here?

spec/API_specification/dataframe_api/__init__.py
MarcoGorelli: what do you suggest?

spec/API_specification/dataframe_api/__init__.py
MarcoGorelli: yup I've removed this thanks

spec/API_specification/dataframe_api/__init__.py
kkraus14: If there's an expectation that it's represented by a 64 bit integer under the covers (and the limitations that imposes)?

MarcoGorelli: not really sure what to say is supported, as chrono is currently limited to dates in the -280,000, 280,000 range https://docs.rs/chrono/latest/chrono/#limitations yet the `'ms'` range should be able to go beyond that

maybe just leave it as an implementation detail for now? else what do you suggest writing?

MarcoGorelli:
I've removed `Date` (as pandas doesn't support it yet), and noted the lack of guarantees for `Datetime`

OK to get this in?

MarcoGorelli:
thanks - shall we keep this out for now so we can tag a first version?

rgommers:
Ah okay, that's why you marked it as draft. Sure, sounds good.

spec/API_specification/dataframe_api/__init__.py
```diff
@@ -238,4 +241,27 @@ class Float32:
...
-    """Boolean type with 8 bits of precision."""
+    """Boolean type."""
+
+class Datetime:
+    """
+    Datetime type.
+
+    Attributes
+    ----------
+    time_unit : Literal['ms', 'us', 'ns']
```
kkraus14, line 252: Having gone through some work in this area recently, nanosecond support isn't common in relational databases where for example DuckDB and Postgres do datetimes as number of microseconds ([DuckDB](https://duckdb.org/docs/sql/data_types/timestamp) [Postgres](https://www.postgresql.org/docs/current/datatype-datetime.html)). Going from microseconds upwards to milliseconds and seconds is generally more easily handled and straightforward, but going down to nanoseconds is not.

Maybe we should consider punting on nanoseconds for the time being?

MarcoGorelli: sure

rgommers (approved):
Mypy is unhappy with the use of `|` instead of `Union` - syntax too recent.

Other than that, LGTM now and the last concern about nanoseconds was addressed by removing `'ns'`. Touched upon again in the last call, and seems good to move ahead. Can be merged once CI is green.

MarcoGorelli:
`from __future__ import annotations` is all we need

cool, thanks, merging then

spec/API_specification/dataframe_api/dtypes.py
```diff
@@ -31,3 +36,35 @@ class Float32:
 class Bool:
     """Boolean type with 8 bits of precision."""
 
+class Date:
+    """
+    Date type.
+
+    There is no guarantee about the range of dates available.
+    """
```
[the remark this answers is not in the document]
jorisvandenbossche: Whoops, sorry ignore my comment, I thought I was commenting on a PR for the interchange protocol, but this _is_ for the standard API ;) So here of course only behaviour matters, and the underlying buffers are an implementation detail of the library.

Was confused because we linked to this from a discussion about supporting "date" in the interchange protocol in the pyarrow implementation.

PR #198, opened 2023-07-12
Rename dtype typevar

### LINKED ISSUE #189: DType parent class / type alias
Also splitting this off from https://github.com/data-apis/dataframe-api/pull/187#discussion_r1245461277

We have various dtypes: `Int64`, `Int32`, ...., `Bool`, which are part of the API.

We have functions which accept any dtype, such as
```python
def column_from_sequence(sequence: Sequence[Scalar], *, dtype) -> Column:
```

How do we type `dtype`? I originally wanted to have `DType` be a common superclass of the dtypes, so we could have
```python
def column_from_sequence(sequence: Sequence[Scalar], *, dtype: DType) -> Column:
```
but it's been brought up that this isn't desirable because it would add a class to the namespace for the sole purpose of type checking.

Are there alternatives which would be acceptable?

MarcoGorelli:
typevars usually end with `T` to distinguish them from type aliases : https://pylint.pycqa.org/en/latest/user_guide/messages/convention/invalid-name.html

I'd suggest we make the typevar `DTypeT`, and then introduce a type alias `DType`, which would resolve #189


rgommers:
> I'd suggest we make the typevar `DTypeT`, and then introduce a type alias `DType`, which would resolve #189

That sounds reasonable to me. Wouldn't you want to introduce the alias in this same PR, so that the signatures and how they're rendered don't change?

rgommers:
Hmm, that didn't completely work, maybe because of the `if TYPE_CHECKING`. In the preview for this PR we have for `Column.get_rows` (as an example):
```
get_rows(indices: Column[Any]) → Column[DtypeT]
```
so compared to in `main`, `Dtype` changed to `DtypeT`: [Column.get_rows](https://data-apis.org/dataframe-api/draft/API_specification/column_object.html#dataframe_api.Column.get_rows)

MarcoGorelli:
isn't that one correct? `DTypeT` is the typevar, so that's what we need to use in `Column[`

image

rgommers:
Oh wait, I was looking at a partial diff, you actually did change the `get_rows` signature in this PR still. I thought the point of our discussion on type aliases was to revert that, and only use `DType`. A "typevar" does not belong in any html rendering, it's an irrelevant type-checking detail to the reader.

MarcoGorelli:
happy to agree to disagree on this one, closing then

PR #199, opened 2023-07-12
Add an `is_dtype` function

MarcoGorelli:


spec/API_specification/dataframe_api/__init__.py
```diff
@@ -177,3 +178,36 @@ class Float32:
...
+        kind: str
+            data type kind.
+            The function must return a boolean indicating whether
+            the input dtype is of a specified data type kind.
+            The following dtype kinds must be supported:
+                - 'bool': boolean data type (Bool).
+                - 'signed integer': signed integer data types (Int8, Int16, Int32, Int64).
+                - 'unsigned integer': unsigned integer data types (UInt8, UInt16, UInt32, UInt64).
+                - 'temporal': temporal data types (Date, Datetime, Duration).
+                - 'string': string data type (String).
```
MarcoGorelli, line 200: we don't have these yet, so https://github.com/data-apis/dataframe-api/pull/197 should ideally go in first

spec/API_specification/dataframe_api/__init__.py
```diff
@@ -177,3 +178,36 @@ class Float32:
...
+
+def is_dtype(dtype: Any, kind: str | tuple[str, ...]) -> bool:
+    """
+    Returns a boolean indicating whether a provided dtype is of a specified data type “kind”.
+
+    Parameters
+    ----------
+        dtype: Any
+            The input dtype.
+        kind: str
```
MarcoGorelli, line 191: the array api also allows `dtype`, but I don't think we need that here. If you want to check whether a dtype is of a specific dtype, you can just use `isinstance`, e.g.
```python
isinstance(column.dtype, namespace.Float64)
```

I think this function is more useful for checking conceptual hierarchies, like `'integral'`

kgryte: Part of the reason for allowing dtype instances is to allow arbitrary composition of dtype sets beyond the supported kinds and allow an escape hatch beyond the kinds officially supported by the array API standard.

MarcoGorelli: can't you do that with
```python
is_dtype(column.dtype, 'integral') or isinstance(column.dtype, MyFancyDType)
```
?

kgryte: Yes, but that becomes a bit unwieldy quickly if you want to arbitrarily combine multiple "kinds" and/or individual dtypes.

MarcoGorelli: I'd suggest keeping this simple to begin with, we can always increase complexity later if necessary

rgommers (approved):
LGTM to merge - it's not complete, as discussed in the open comment thread, but this works as is and the rest can be done later. 

Thanks @MarcoGorelli and @kgryte

PR #200, opened 2023-07-12
Add column.slice_rows and column.get_rows_by_mask

MarcoGorelli:
Here's another couple of dataframe methods missing for column

### DIFF
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -92,6 +92,46 @@ class Column(Generic[DType]):
         """
         ...
 
+
+    def slice_rows(
+        self: Column[DType], start: int | None, stop: int | None, step: int | None
+    ) -> Column[DType]:
+        """
+        Select a subset of rows corresponding to a slice.
+
+        Parameters
+        ----------
+        start : int or None
+        stop : int or None
+        step : int or None
+
+        Returns
+        -------
+        Column
+        """
+        ...
+
+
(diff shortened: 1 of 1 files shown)


kkraus14 (approved):

spec/API_specification/dataframe_api/column_object.py
```diff
@@ -92,6 +92,46 @@ def get_rows(self: Column[DType], indices: Column[Any]) -> Column[DType]:
...
+        mask : Column[bool]
+
+        Returns
+        -------
+        Column
+
+        Notes
+        -----
+        Some participants preferred a weaker type Arraylike[bool] for mask,
+        where 'Arraylike' denotes an object adhering to the Array API standard.
```
rgommers, line 130: I think it's fine to leave this note here for now, useful to capture the differing opinions while moving forward. We can remove it once we polish things for a final release.

spec/API_specification/dataframe_api/column_object.py
```diff
@@ -92,6 +92,46 @@ def get_rows(self: Column[DType], indices: Column[Any]) -> Column[DType]:
         """
         ...
 
+
+    def slice_rows(
+        self: Column[DType], start: int | None, stop: int | None, step: int | None
```
rgommers, line 97: Allowing both positional and keyword usage seems good to me here.

rgommers (approved):
LGTM, thanks Marco.

spec/API_specification/dataframe_api/column_object.py
MarcoGorelli: thanks - I just copied it from the dataframe counterpart

PR #206, opened 2023-07-24
Move api_version to dataframe_standard

### LINKED ISSUE #205: Require `api_version` argument in `__dataframe_standard__` rather than `__dataframe_namespace__`?
## Reminder: what these magic methods are

Currently, the way to convert a non-compliant dataframe to a compliant one is by calling `df.__dataframe_standard__`:

https://github.com/data-apis/dataframe-api/blob/16dea0b4f3b5d732f71d85529fb42242f8669962/spec/purpose_and_scope.md#L362-L365

Once you've got a compliant dataframe, you can get the namespace with `__dataframe_namespace__`

https://github.com/data-apis/dataframe-api/blob/3111499ae1169d83f97460afdb91141a2721346b/spec/API_specification/dataframe_api/dataframe_object.py#L40-L54

## Why `__dataframe_standard__` needs `api_version`:

Take the following example:
```python
def remove_outliers(df, column):
    # Get a Standard-compliant dataframe.
    df_standard = df.__dataframe_standard__(api_version="2023.07")
    # Use methods from the Standard specification.
    col = df_standard.get_column_by_name(column)
    z_score = (col - col.mean()) / col.std()
    df_standard_filtered = df_standard.get_rows_by_mask((z_score > -3) & (z_score < 3))
    # Return the result as a dataframe from the original library.
    return df_standard_filtered.dataframe
```

I'm not using `__dataframe_namespace__` here, so the only way I have of asking for a certain api_version of the standard is via `__dataframe_standard__`

## Why `__dataframe_namespace__` probably doesn't need `api_version`

Say I do
```python
df_standard = df.__dataframe_standard__(api_version="2023.07")
namespace = df_standard.__dataframe_namespace__()
```
Then it seems natural that the namespace returned would be the "2023.07" one. So it doesn't need repeating in `__dataframe_namespace__`

MarcoGorelli:
closes #205

### DIFF
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -37,21 +37,9 @@ class DataFrame:
     **Methods and Attributes**
 
     """
-    def __dataframe_namespace__(
-        self, /, *, api_version: str | None = None
-    ) -> Any:
+    def __dataframe_namespace__(self) -> Any:
         """
-        Returns an object that has all the dataframe API functions on it.
-
-        Parameters
-        ----------
-        api_version: Optional[str]
-            String representing the version of the dataframe API specification
-            to be returned, in ``'YYYY.MM'`` form, for example, ``'2023.04'``.
-            If it is ``None``, it should return the namespace corresponding to
-            latest version of the dataframe API specification.  If the given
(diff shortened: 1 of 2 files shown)


kkraus14 (approved):

MarcoGorelli:
thanks, for the approval, merging then

PR #208, opened 2023-07-24
Add 2d constructor

MarcoGorelli:
Also a request from scikit-learn

### DIFF
--- a/spec/API_specification/dataframe_api/__init__.py
+++ b/spec/API_specification/dataframe_api/__init__.py
@@ -96,6 +96,20 @@ def dataframe_from_dict(data: Mapping[str, Column[Any]]) -> DataFrame:
     """
     ...
 
+def dataframe_from_2d_array(array: Any) -> DataFrame:
+    """
+    Construct DataFrame from 2D array.
+
+    See `column_from_sequece` for related 1D function.
+
+    Only Array-API-compliant 2D arrays are supported.
+
+    Returns
+    -------
+    DataFrame
+    """
+    ...
+
 class null:
     """
     A `null` object to represent missing data.
--- a/spec/API_specification/dataframe_api/__init__.py
+++ b/spec/API_specification/dataframe_api/__init__.py
@@ -96,13 +96,22 @@ def dataframe_from_dict(data: Mapping[str, Column[Any]]) -> DataFrame:
     """
     ...
 
-def dataframe_from_2d_array(array: Any) -> DataFrame:
+def dataframe_from_2d_array(array: Any, *, names: Sequence[str], dtypes: Mapping[str, Any]) -> DataFrame:
     """
     Construct DataFrame from 2D array.
 
-    See `column_from_sequece` for related 1D function.
+    See `column_from_sequence` for related 1D function.
 
     Only Array-API-compliant 2D arrays are supported.
+
+    Parameters
+    ----------
+    array : array
+        array-API compliant 2D array
+    names : Sequence[str]
+        Names to give columns. Must be the same length as ``array.shape[1]``.
+    dtypes : Mapping[str, DType]
+        Dtype of each column. Must be the same length as ``array.shape[1]``.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


spec/API_specification/dataframe_api/__init__.py
```diff
@@ -96,6 +96,20 @@ def dataframe_from_dict(data: Mapping[str, Column[Any]]) -> DataFrame:
     """
     ...
 
+def dataframe_from_2d_array(array: Any) -> DataFrame:
```
rgommers, line 99: This should have `*, names: list[str]` I think, with the length of the list matching the number of columns?

Also a `dtype` keyword, because we don't really want to get in the business of inferring it (going from `array.dtype` to a dtype in this namespace is nontrivial; much easier for the caller who has the array namespace at hand).

spec/API_specification/dataframe_api/__init__.py
```diff
@@ -96,6 +96,20 @@ def dataframe_from_dict(data: Mapping[str, Column[Any]]) -> DataFrame:
...
+    """
+    Construct DataFrame from 2D array.
+
+    See `column_from_sequece` for related 1D function.
```
rgommers, line 103: typo in `sequence`

rgommers: More importantly, I foresee a `column_from_1d_array` request, because a 1-D array is not a sequence.

rgommers (commented):
This seems like a reasonable addition to me, especially given that scikit-learn would need it and we want a dataframe standard <--> array standard bridge.

spec/API_specification/dataframe_api/__init__.py
MarcoGorelli: I wasn't aware, thanks

just for my understanding, what makes it not a sequence? from the docs https://docs.python.org/3/glossary.html#term-sequence :

> An [iterable](https://docs.python.org/3/glossary.html#term-iterable) which supports efficient element access using integer indices via the __getitem__() special method and defines a __len__() method that returns the length of the sequence

, don't arrays meet the requirement?

rgommers: See https://github.com/data-apis/array-api/issues/481 for `len()`. A second issue is that iterating over it is going to return 0-D arrays of the dtype of the input array, not Python scalars.

kkraus14: Are arrays iterable per the array API? I would strongly prefer to differentiate array handling vs iterable handling since you can often handle array zero copy or at least in very efficient ways relative to iterables.

MarcoGorelli: sure, I've added column from 1d too

spec/API_specification/dataframe_api/__init__.py
```diff
@@ -96,6 +98,53 @@ def dataframe_from_dict(data: Mapping[str, Column[Any]]) -> DataFrame:
...
+    Only Array-API-compliant 2D arrays are supported.
+
+    Parameters
+    ----------
+    array : array
+        array-API compliant 2D array
+    names : Sequence[str]
+        Names to give columns. Must be the same length as ``array.shape[1]``.
+    dtypes : Mapping[str, DType]
+        Dtype of each column. Must be the same length as ``array.shape[1]``.
```
kkraus14, line 140: What type of casting do we allow vs disallow here? I.E. if someone passes an array of type `float32` and a mapping that uses `bool` dtypes here, is that something we expect to support?

MarcoGorelli: No cross-kind casting? So, float32 to bool would be disallowed?

kkraus14: Do we allow downcasting within kind or only upcasting? I.e. is float64 array --> float32 column allowed? Or only float32 array --> float32 column?

MarcoGorelli: is there any array-api precedent to go off of? (haven't looked to carefully yet)

rgommers: There is. Here are the casting rules: https://data-apis.org/array-api/latest/API_specification/type_promotion.html#type-promotion. In summary:
- cross-kind casting is undefined, given that not all libraries allow that
- downcasting is disallow in general, and if you're sure that that is what you want, there is a single function (`astype`) to explicitly downcast

MarcoGorelli: thanks - OK to just go with that?

PR #209, opened 2023-07-26
Add DataFrame.to_array_object and Column.to_array_object

MarcoGorelli:
Trying to move this forwards, according to my understanding

Sorry if I've misunderstood the

>         While numpy arrays are not yet array-API-compliant, implementations
>        may choose to return a numpy array (for numpy prior to 2.0), with the
>        understanding that consuming libraries would then use the
>        ``array-api-compat`` package to convert it to a Standard-compliant array.

part

### DIFF
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -700,3 +700,25 @@ class Column(Generic[DType]):
 
         """
         ...
+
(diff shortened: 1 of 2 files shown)


kkraus14 (approved):

MarcoGorelli:
thanks for your review! merging then as there haven't been objections

rgommers:
I had a post-hoc look - this LGTM, thanks both!

PR #211, opened 2023-07-26
let `keys` be none in sorted_indices

MarcoGorelli:
a few other methods allow this (`unique_indices`, `fill_null`) so we should probably add it here too

### DIFF
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -253,7 +253,7 @@ class DataFrame:
 
     def sorted_indices(
         self,
-        keys: Sequence[str],
+        keys: Sequence[str] | None = None,
         *,
         ascending: Sequence[bool] | bool = True,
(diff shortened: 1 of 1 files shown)


spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -267,8 +267,9 @@ def sorted_indices(
 
         Parameters
         ----------
-        keys : Sequence[str]
+        keys : Sequence[str] | None
             Names of columns to sort by.
+            If `None`, sort by all columns.
```
kkraus14, line 272: I don't think we've explicitly stated or clarified anywhere that columns are ordered within DataFrames and that there is a concept of a 0th or 1st column. Maybe we should?

kkraus14 (approved):

spec/API_specification/dataframe_api/dataframe_object.py
MarcoGorelli: sure, no objections if you want to add it

PR #213, opened 2023-08-01
to_array_object: type dtype as str, note possibilities

MarcoGorelli:
not all dataframe dtypes are supported by the array api (e.g. `datetime`), so just accepting the same ones and sometimes raising `NotImplementedError` may not be the best experience

How about just accepting `str`, and listing the allowed strings? These dtypes don't need to store anything (like categories / time unit / time zone), so don't need to be classes anyway.
Internally, each implementation can map to the correct object (e.g. for pandas: 'int64' would go to np.int64)

### DIFF
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -699,17 +699,17 @@ class Column(Generic[DType]):
             The dtype of the array-API-compliant object to return.
             Must be one of:
 
-            - namespace.Bool()
-            - namespace.Int8()
-            - namespace.Int16()
-            - namespace.Int32()
-            - namespace.Int64()
-            - namespace.UInt8()
-            - namespace.UInt16()
-            - namespace.UInt32()
-            - namespace.UInt64()
-            - namespace.Float32()
-            - namespace.Float64()
+            - Bool()
+            - Int8()
+            - Int16()
+            - Int32()
+            - Int64()
+            - UInt8()
+            - UInt16()
+            - UInt32()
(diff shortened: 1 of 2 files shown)


rgommers (commented):
Listing the accepted dtypes seems useful, but I don't see why we'd want it to be strings rather than dtypes. The latter is much more natural, and it must work because this is still within a single library hence all dtype objects/instances are known entities.

MarcoGorelli:
sure sounds good

spec/API_specification/dataframe_api/column_object.py
```diff
@@ -695,8 +695,21 @@ def to_array_object(self, dtype: Any) -> Any:
 
         Parameters
         ----------
-        dtype : Any
+        dtype : DType
             The dtype of the array-API-compliant object to return.
+            Must be one of:
+
+            - namespace.Bool()
```
rgommers, line 702: the `namespace.` seems confusing to me, and I don't think we've used that anywhere else. One of these should work:
```
Must be an instance of 

- `Bool`
- ` Int8`
- etc
```
or, more concisely:
```
Must be a boolean or numerical (signed or unsigned integer, or floating-point) dtype that is part of this standard.
```

MarcoGorelli: thanks - sure I've removed `namespace`

do we want to keep `Bool()` to remind that it's a class they need to pass?

rgommers (approved):

MarcoGorelli:
thanks for your review!

PR #214, opened 2023-08-01
Rename entrypoints to `__dataframe_consortium_standard__` and `__column_consortium_standard__`

MarcoGorelli:
Renaming to make this more amenable to pandas


rgommers (approved):
Sure. It's not a good name, but whatever makes folks happy here so we can move ahead.

PR #215, opened 2023-08-01
Fix up some return types for methods with `skip_nulls`

MarcoGorelli:
Some column methods can return `null` if `skip_nulls` is `False`, but the return annotations don't currently capture that

~edit: wait, `null` isn't a valid type~ have added `NullType` (but just used internally, it's not exported)


rgommers (approved):
Thanks Marco! This seems to work and be quite clear in the docs still, so in it goes.

I had to think a bit about why `null` isn't valid. I guess since it's a scalar with a given name rather than a class of which we return instances, the alternative here would have been `Literal[null]` - but that's not any clearer, so I'm happy with `NullType`.

PR #216, opened 2023-08-01
remove api_version from column namespace

MarcoGorelli:
forgot this one in https://github.com/data-apis/dataframe-api/pull/206


kkraus14 (approved):

rgommers (approved):
This is consistent with gh-206, so LGTM from that perspective. I'm not convinced that we won't also need versioning in the `__*_namespace__` methods, but if so we'll bump into that in the future and can easily add it back there. So in it goes. Thanks @MarcoGorelli

PR #219, opened 2023-08-02
Add dataframe.collect

### LINKED ISSUE #120: DataFrame.collect() for lazy dataframes?
Some DataFrames have a lazy api, which the standard should probably support

Should we add a `collect` method, which for eager libraries would just return `self` and for lazy ones would materialise the dataframe?

MarcoGorelli:
this has come up in a few discussions

closes #120

### DIFF
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -838,3 +838,16 @@ class DataFrame:
         understanding that consuming libraries would then use the
         ``array-api-compat`` package to convert it to a Standard-compliant array.
         """
+
+    def collect(self) -> Any:
+        """
+        Materialize the DataFrame.
+
+        Non-lazy implementations may just return `self`.
+
+        Returns
+        -------
+        Any
+            The materialized object.
+        """
+        ...


spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -838,3 +838,16 @@ def to_array_object(self, dtype: Any) -> Any:
         understanding that consuming libraries would then use the
         ``array-api-compat`` package to convert it to a Standard-compliant array.
         """
+
+    def collect(self) -> Any:
```
shwina, line 842: Existing APIs of lazy dataframes look quite different and the equivalent of the `collect()` function accepts various implementation-specific kwargs [1], [2].

Can it be assumed that each library has "reasonable defaults" that will work for most use-cases, and those can be used in their implementations of `collect()`? (I think it can, but just making sure we consider this).

[1]: https://docs.dask.org/en/stable/generated/dask.dataframe.compute.html
[2]: https://pola-rs.github.io/polars/py-polars/html/reference/lazyframe/api/polars.LazyFrame.collect.html

rgommers (changes requested):
Can we keep the discussion on gh-120? I don't see anything that has changed here, conceptually this does not seem like a good idea. Will comment more on gh-120.

PR #221, opened 2023-08-02
add column.rename

MarcoGorelli:
columns can have names, so they probably need a `rename` method to change it?


rgommers:
Should this explicitly say that this returns a new column that's not contained in any dataframe? I imagine a lot of folks are going to assume it's an in-place change, which cannot really work.

rgommers (approved):

kkraus14 (approved):

PR #222, opened 2023-08-02
dataframe_from_2d_array: only accept single dtype

MarcoGorelli:
I'd over-complicated this one: the array only has a single dtype, so there only needs to be a single `dtype` argument


kkraus14:
I thought that we decided we were going to follow the array typecasting rules which is roughly that upcasting within the same type is allowed? Say I have an `int8` array, this means I could reasonably say give me a DataFrame with `int8`, `int16`, `int32`, and `int64` columns?

PR #223, opened 2023-08-02
Don't let Column.name be None

MarcoGorelli:
Noticed this while updating the polars implementation: their series must have a `str` name, and can't be `None`

We _could_ hack around this in the implemenation, but I think it's easier to just require the name to only be `str`

Afterall:
- if a column is created from a dataframe (e.g. `df.get_column_by_name('foo')`), then its name will be the column name (which has to be `str`)
- if a column is constructed from a sequence or 1d array, then we can require that a `str` `name` be specified

So, `name` will always be `str`

### DIFF
--- a/spec/API_specification/dataframe_api/__init__.py
+++ b/spec/API_specification/dataframe_api/__init__.py
@@ -72,7 +72,7 @@ def column_from_sequence(sequence: Sequence[Any], *, dtype: Any, name: str = '')
         Sequence of elements. Each element must be of the specified
         ``dtype``, the corresponding Python builtin scalar type, or
         coercible to that Python scalar type.
-    name : str
+    name : str, optional (default: '')
         Name of column.
     dtype : DType
         Dtype of result. Must be specified.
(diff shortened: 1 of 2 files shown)


spec/API_specification/dataframe_api/__init__.py
```diff
@@ -72,7 +72,7 @@ def column_from_sequence(sequence: Sequence[Any], *, name: str | None, dtype: An
         Sequence of elements. Each element must be of the specified
         ``dtype``, the corresponding Python builtin scalar type, or
         coercible to that Python scalar type.
-    name : str, optional
+    name : str
```
kkraus14, line 75: Instead of making name a required field, could we default it to an empty string if we want to disallow `None`? There's a lot of column only use cases where there isn't really a used name of the column and this would just add noise.

MarcoGorelli: ok sure

kkraus14: This is now `optional` again I believe?

kkraus14 (approved):

PR #226, opened 2023-08-07
Remove `nan_as_null` parameter for DataFrame protocol

### LINKED ISSUE #125: Interchange Protocol: usused `nan_as_null` keyword?
The DataFrame Interchange Protocol has a `nan_as_null` keyword in `__dataframe__` that can be specified by the consumer, i.e. the person/library calling this method. The docstring explains its goal:

https://github.com/data-apis/dataframe-api/blob/d10a096ccc1612ec4acf8503aff58b3acb4e3738/protocol/dataframe_protocol.py#L400-L403

However, at the moment I think none of the existing implementation we are aware of actually supports this keyword: pandas, vaex, modin, cudf all simply ignore it (and cudf actually uses a wrong default), and pyarrow as well but will raise an error if you pass a non-default value (i.e. True). 
The following disclaimer in the docstring seems to have been copied around as some version of this appears in all the listed projects:

```
        `nan_as_null` currently has no effect; once support for nullable extension
        dtypes is added, this value should be propagated to columns.
```

I am not fully sure what this note is meant for (I suppose it originates from the pandas implementation where for nullable extension types this keyword could be implemented? Since the default numpy dtypes already use NaN as null, the keyword wouldn't have any effect for that. But for other libraries that explanation doesn't necessarily hold).  
And also, if nobody implemented it, do we actually need it?

stinodego:
Closes #125

We [discussed](https://github.com/pola-rs/polars/pull/10267#issuecomment-1667595401) this today, figured I'd make a PR!

### DIFF
--- a/protocol/dataframe_protocol.py
+++ b/protocol/dataframe_protocol.py
@@ -392,15 +392,11 @@ class DataFrame(ABC):
 
     @abstractmethod
     def __dataframe__(
-        self, nan_as_null: bool = False, allow_copy: bool = True
+        self, allow_copy: bool = True
     ) -> "DataFrame":
         """
         Construct a new exchange object, potentially changing the parameters.
 
-        ``nan_as_null`` is a keyword intended for the consumer to tell the
-        producer to overwrite null values in the data with ``NaN``.
-        It is intended for cases where the consumer does not support the bit
-        mask or byte mask that is the producer's native representation.
         ``allow_copy`` is a keyword that defines whether or not the library is
         allowed to make a copy of the data. For example, copying data would be
         necessary if a library supports strided buffers, given that this protocol


kkraus14 (approved):

stinodego:
This could be considered a breaking change - not sure if this can just be merged?

rgommers (commented):
> This could be considered a breaking change - not sure if this can just be merged?

I think it can be, as long as no implementation uses it yet. Which is covered largely by the discussion in gh-125 and in https://github.com/pola-rs/polars/pull/10267 I believe. 

However, to make extra sure, let's open PRs that remove the keyword where needed to implementers or ping the relevant authors. So far we covered:
- [x] pandas
- [x] Polars
- [x] PyArrow

In addition, we know of:
- [ ] Modin (Cc @anmyachev as author of https://github.com/modin-project/modin/pull/6361)
- [ ] Vaex (implementation at https://github.com/vaexio/vaex/blob/15245cf4332d4423ac58bd737aee27d911a1b252/packages/vaex-core/vaex/dataframe_protocol.py#L720)
- [ ] static-frame (implementation at https://github.com/static-frame/static-frame/blob/159039d4ec05001990393142cca7928ca04a011a/static_frame/core/protocol_dfi.py#L276)

They all have the keyword in their signatures, so even if it raises when a non-default value is passed in, it can't just be removed from the signature of `__dataframe__` in the implementations, or it's going to break other libraries calling it with `nan_as_null=nan_as_null`.

So perhaps we change the description in the docs here first to "DO NOT USE", then remove it from the consumers calling `__dataframe__`, and only later on remove it from the signature of `__dataframe__`?

stinodego:
> So perhaps we change the description in the docs here first to "DO NOT USE", then remove it from the consumers calling __dataframe__, and only later on remove it from the signature of __dataframe__?

Sounds like the right approach. I just opened a PR with a warning - feel free to edit the text however you fit:
https://github.com/data-apis/dataframe-api/pull/228

rgommers:
With gh-228 merged, let's close this PR. Thanks @stinodego!

stinodego:
We'll still need to remove the parameter once the consumers have been updated, but up to you!

rgommers:
Yes for sure - we can reopen this one later. It's just easier to not have PRs in the queue that aren't mergeable for quite a while.

PR #230, opened 2023-08-21
Rename insert, remove label argument

MarcoGorelli:
There's a couple of things I don't like about `DataFrame.insert`:
- the name is less descriptive than the rest of the API
- the argument `label` isn't really needed, because we have both `Column.name` and `Column.rename`


kkraus14 (approved):
LGTM

MarcoGorelli:
thanks all

Issue #229, opened 2023-08-21
Should there be `namespace.col` for filtering?


MarcoGorelli:
## TL;DR

Add `namespace.col`, to allow for lazy columns / lazy column reductions to work seamlessly in lazy implementations

Then:
- use `col = df.get_column_by_name('a')` if you want an eager column 'a', on which you can call reductions and comparisons (like `col.mean()`,  `col > 0`). Not necessarily available for all implementations (e.g. Ibis)
- use `namespace.col('a')` if you want a (possibly lazy) expression which you intend to use for filtering, like `df.get_rows_by_mask((namespace.col('a') - namespace.col('a').mean()) > 0)`

## Longer proposal

I'm a bit concerned about some operations just not being available in lazy cases in general, such as
```python
mask = df2.get_column_by_name('a') > 0
df1.get_rows_by_mask(mask)
```

~Polars~ Ibis at least doesn't allow this, filtering by a mask is only allowed if the mask is a column which comes from the same dataframe that's being filtered on - so
```python
mask = df1.get_column_by_name('a') > 0
df1.get_rows_by_mask(mask)
```
would be allowed instead. ~Such comparisons are intentionally not implemented in polars, see https://github.com/pola-rs/polars/issues/10274 for a related issue (`__eq__` between two different dataframes)~

I'd like to suggest, therefore, that we introduce `namespace.col`, so that the above could become
```python
mask = namespace.col('a') > 0
df1.get_rows_by_mask(mask)
```

And then there doesn't need to be any documentation like "this won't work for some implementations if the mask was created from a different dataframe"

This would also feel more familiar to developers:
- `pyspark` already has this:
  ```python
  from pyspark.sql.functions import col
  df.filter(col("state") == "OH")
  ```
- `polars` already has this (same as above, but `from polars import col`)
- `pandas` _may_ add it (see EuroScipy 2023 lightning talk by Joris)
- ibis has `_`
- siuba also has `_`

EDIT: have tried clarifying the issue. Have also replaced Polars with Ibis in the example as the Consortium seems more interested in that

kkraus14:
> Polars at least doesn't allow this. In the implementation I've put together, filtering by a mask is only allowed if the mask is a column which comes from the same dataframe that's being filtered on

This feels like an arbitrary limitation of Polars that I haven't seen in other implementations where my inclination would be to say that this should be addressed in Polars as opposed to having the limitation built into the standard.

[...]

PR #231, opened 2023-08-22
Add DataFrame.insert_columns

MarcoGorelli:
Example from polars of where this makes a difference:

Inserting sequentially:
```python
In [19]: df = pl.LazyFrame({"a": [1, 1, 2], 'b': [4,5,6]})

In [20]: df = df.with_columns(a_plus_one=pl.col('a')+1)

In [21]: df = df.with_columns(b_plus_one=pl.col('b')+1)

In [22]: print(df.explain())
 WITH_COLUMNS:
 [[(col("b")) + (1)].alias("b_plus_one")]
   WITH_COLUMNS:
   [[(col("a")) + (1)].alias("a_plus_one")]
    DF ["a", "b"]; PROJECT */2 COLUMNS; SELECTION: "None"
```

Inserting in parallel, as `a_plus_1` and `b_plus_1` are independent:
```python
In [25]: df = pl.LazyFrame({"a": [1, 1, 2], 'b': [4,5,6]})

In [26]: df = df.with_columns(
    ...:     a_plus_one=pl.col('a')+1,
    ...:     b_plus_one=pl.col('b')+1,
    ...: )

In [27]: print(df.explain())
 WITH_COLUMNS:
 [[(col("a")) + (1)].alias("a_plus_one"), [(col("b")) + (1)].alias("b_plus_one")]
  DF ["a", "b"]; PROJECT */2 COLUMNS; SELECTION: "None"
```

### DIFF
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -219,7 +219,8 @@ class DataFrame:
         Parameters
         ----------
         columns : Column | Sequence[Column]
-            Column(s) to insert. Insertion order is not guaranteed.
+            Column(s) to insert. These will appear at the rightmost location, but
+            insertion order is not guaranteed.
         """
         ...
 
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -209,6 +209,52 @@ class DataFrame:
         """
         ...
 
+    def insert_columns(self, columns: Sequence[Column[Any]]) -> DataFrame:
+        """
+        Insert columns into DataFrame at rightmost location.
+
+        Like :meth:`insert_column`, but can insert multiple (independent) columns.
+        Some implementations may be able to make use of parallelism in this
+        case. For example instead of:
+        
+        .. code-block::
+
+            new_column = df.get_column_by_name('a') + 1
+            df = df.insert_column(new_column.rename('a_plus_1'))
+            new_column = df.get_column_by_name('b') + 1
+            df = df.insert_column(new_column.rename('b_plus_1'))
+        
+        it would be better to write
+
+        .. code-block::
+
+            new_column_0 = df.get_column_by_name('a') + 1
+            new_column_1 = df.get_column_by_name('b') + 1
+            df = df.insert_columns(
+                [
+                    new_column_0.rename('a_plus_1'),
+                    new_column_1.rename('b_plus_1'),
+                ]
+            )
+        
+        so that insertion can happen in parallel for some implementations.
+
+        Parameters
+        ----------
+        columns : Sequence[Column]
+            Sequence of `Column`s.
+            Must be independent of each other.
+            Column names must be unique.
+            Column names may not already be present in the
+            dataframe - use :meth:`Column.rename` to rename them
+            beforehand if necessary.
+        
+        Returns
+        -------
+        DataFrame
+        """
+        ...
+
     def drop_column(self, label: str) -> DataFrame:
         """
         Drop the specified column.
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -206,7 +206,7 @@ class DataFrame:
         If inserting multiple columns, they must be indepedent.
         For example, instead of
         
-        .. code-block::
+        .. code-block::python
 
             new_column = df.get_column_by_name('a') + 1
             df = df.insert_column(new_column.rename('a_plus_1'))
@@ -215,7 +215,7 @@ class DataFrame:
         
         it would be better to write
 
-        .. code-block::
+        .. code-block::python
 
             new_column_0 = df.get_column_by_name('a') + 1
             new_column_1 = df.get_column_by_name('b') + 1
@@ -233,7 +233,7 @@ class DataFrame:
         columns : Column | Sequence[Column]
             Column(s) to insert. Must be independent of each other. For example,
 
-            .. code-block:: python
+            .. code-block::python
                 new_column_1 = df.get_column_by_name('a').rename('b')
                 new_column_2 = (new_column_1 + 2).rename('c')
                 df.insert_columns([new_column_1, new_column_2])
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -180,9 +180,9 @@ class DataFrame:
         """
         ...
 
-    def insert_column(self, column: Column[Any]) -> DataFrame:
+    def insert_columns(self, columns: Column[Any] | Sequence[Column[Any]]) -> DataFrame:
         """
-        Insert column into DataFrame at rightmost location.
+        Insert column(s) into DataFrame at rightmost location.
 
         The column's name will be used as the label in the resulting dataframe.
         To insert the column with a different name, combine with `Column.rename`,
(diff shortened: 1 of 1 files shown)


MarcoGorelli:
needs rebasing onto #239

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -180,27 +180,81 @@ def get_rows_by_mask(self, mask: Column[Bool]) -> DataFrame:
...
 
         Parameters
         ----------
-        loc : int
-            Insertion index. Must verify 0 <= loc <= len(columns).
         column : Column
         """
         ...
 
+    def insert_columns(self, columns: Sequence[Column[Any]]) -> DataFrame:
```
kkraus14, line 212: Could we have just this API and remove the `insert_column` API to only have a single API supported? We could then make `columns` be `Column[Any] | Sequence[Column[Any]]`?

MarcoGorelli: yup, that's a really good point, thanks!

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -203,9 +203,22 @@ def insert_column(self, column: Column[Any]) -> DataFrame:
             df = df.insert_column(new_column.rename('a_plus_1'))
             df = df.get_columns_by_name(new_column_names)
 
+        If inserting multiple columns, they must be indepedent.
```
rgommers, line 206: minor: typo in independent

rgommers (commented):
Inserting multiple columns at once seems clearly useful to me. The "columns must be independent" is a little concerning though - it seems like a fuzzy boundary and I'm not even sure how independent is defined. Whether two columns are truly independent may be very hard to establish within the context of a single function. What if I have a function like:
```python
def a_func(df, col1, col2):
    return df.insert_columns(col1, col2)
```
and `col1` and `col2` are related through some earlier operation one or more levels up in the call stack?

It seems to me like this restriction should be dropped, and that instead the implementation is responsible for correctly serializing the operations if it detects that the columns are related somehow.

MarcoGorelli:
> and col1 and col2 are related through some earlier operation one or more levels up in the call stack?

If you really don't know, then you can just call `DataFrame.insert` twice:
```python
def a_func(df, col1, col2):
    return df.insert_columns(col1).insert_columns(col2)
```

Anyway, I've rephrased to "insertion order is not guaranteed and may vary across implementation", so if you try inserting one column and then another one which is a transformation of the one you just inserted, then that's not supported

If you do insert multiple columns at the same time, then it's up to you (the user) to know that insertion can happen in an "embarassingly parallel" manner

MarcoGorelli:
> and that instead the implementation is responsible for correctly serializing the operations if it detects that the columns are related somehow.

How, with a `try-except`? I'd rather keep such magic fallbacks out, but thanks for the suggestion

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -203,9 +203,23 @@ def insert_column(self, column: Column[Any]) -> DataFrame:
...
+        following
+
+        .. code-block:: python
+
+            new_column_1 = df.get_column_by_name('a').rename('b')
+            new_column_2 = (new_column_1 + 2).rename('c')
+            df.insert_columns([new_column_1, new_column_2])
+            
+        is not supported, as `new_column_2` is derived from `new_column_1`, which may
+        not be part of the dataframe if `new_column_2` is inserted first.
```
kkraus14, line 217: As of now our `Column` objects are immutable and have no concept of being derived from. Could you explain why this wouldn't be supported?

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -203,9 +203,23 @@ def insert_column(self, column: Column[Any]) -> DataFrame:
...
+
         Parameters
         ----------
-        column : Column
+        columns : Column | Sequence[Column]
+            Column(s) to insert. Insertion order is not guaranteed.
```
kkraus14, line 222: The first line of the docstring we say that columns are inserted at the rightmost location, which I think is the expected behavior that most users would want. Is there something that prevents us from not being able to guarantee insertion order?

spec/API_specification/dataframe_api/dataframe_object.py
MarcoGorelli: Say you start with
```
a b
1 4
1 5
2 6
```
and you want to add two new columns:
- `c`: which is `b` + 1
- `d`: which is `a` + `c`

These two new columns can't be added in any order, they need to be added first 'c' and then 'd'

If you try inserting 'd' before 'c', then you could get "KeyError: column not found, 'c' not part of dataframe"

kkraus14: Okay that makes sense as to why in a lazy implementation you can't use derived columns from other columns being inserted. Assuming that isn't allowed, then can inserting the columns in the order they are passed possible?

MarcoGorelli: eager too, the above example was eager

Should be possible, yes (we could always call `.get_columns_by_name` internally and reoder) I was just thinking about how to get the concept across without saying "independent"

kkraus14: Sorry, by lazy I meant that the columns are expressions as opposed to how we currently have defined `Column`s.

So it sounds like this points back into https://github.com/data-apis/dataframe-api/issues/229 where the current design of `Column` doesn't suffice?

shwina: Sorry for the drive-by comment here, but imposing the restrictions of a lazy implementation on the API in this way seems quite arbitrary. We should either just not support this API (because it cannot be implemented naturally/ efficiently by lazy implementations), or discuss a distinct lazy API.

kkraus14: I'm not sure this is a lazy vs eager thing but moreso a `Column` vs a different kind of object returned by `namespace.col` proposed in #229. The above example given is eager.

Without the `namespace.col` and using `pl.Series` the above example looks something like:
```python
df = pl.DataFrame({'a':[1,1,3], 'b':[4,5,6]})

c = df['a'] + 1
d = c + df['a']

df.with_columns(c=c, d=d)
```

This does not work in Polars lazy mode as it doesn't have `Series` / `Column` objects and instead works via expressions like `namespace.col`.

shwina: The _need_ for a `namespace.col` seems natural for lazy implementations like Polars, but not so much for eager implementations as far as I can tell. So I do think this is a lazy vs eager thing.

I'm getting the sense that conflating lazy and eager into a single API is going to be the source of some tension (and likely a bad UX).

MarcoGorelli: Does Ibis insert the columns in parallel in `select`? If not, I think that's probably the underlying reason

MarcoGorelli:
I've articulated some goals I'd like to see for the project here: https://github.com/data-apis/dataframe-api/issues/244

Currently, inserting multiple independent columns is not a zero-cost-abstraction for the Polars implementation - hence, this PR

I wasn't expecting this to be controversial - let's discuss it on the next call

spec/API_specification/dataframe_api/dataframe_object.py
kkraus14: Ibis generates SQL (or DataFrame code) and hands that generated code to the connected backend, so it's an implementation detail of the backend. But in general, Ibis binds its expressions to Tables which allows writing code like I did above which isn't dependent on a column named "c". On the other hand, the Polars code using expressions you posted isn't bound to a table so it is dependent on having a column named "c".

I think this is again a discussion more about expressions and Columns and what we do in the standard regarding them than this specific function at this point. Should we move discussion back to #229?

MarcoGorelli:
> The need for a namespace.col seems natural for lazy implementations like Polars, but not so much for eager implementations as far as I can tell. So I do think this is a lazy vs eager thing.

It's a readability thing too

Compare
```python
        (
        )
    )
```
with
```python
        (col("sepal_width") > col("sepal_height")) | (col("species") == "setosa")
    )
```

> I'm getting the sense that conflating lazy and eager into a single API is going to be the source of some tension (and likely a bad UX).

Indeed - here's what I'd like to discuss tomorrow https://github.com/data-apis/dataframe-api/pull/249

MarcoGorelli:
closing in favour of #269

PR #234, opened 2023-08-22
Add DataFrame.sort and Column.sort

MarcoGorelli:
Feedback I received when showing this to people is that

```python
col.get_rows(col.sorted_indices())
```
isn't ergonomic

It's also a performance footgun
```python
In [1]: df = pl.DataFrame({'a': np.random.randint(0, 4, size=100_000_000)})

In [2]: %timeit df[df['a'].arg_sort()]
2.56 s ± 80.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [3]: %timeit df.sort('a')
700 ms ± 21.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### DIFF
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -141,6 +141,35 @@ class Column(Generic[DType]):
         """
         ...
 
+    def sort(
+        self,
+        *,
+        ascending: bool = True,
+        nulls_position: Literal['first', 'last'] = 'last',
+    ) -> Column[DType]:
+        """
+        Sort column.
+
+        If you need the indices which would sort the column,
+        use :meth:`sorted_indices`.
+
+        Parameters
+        ----------
+        ascending : bool
+            If `True`, sort in ascending order.
+            If `False`, sort in descending order.
+        nulls_position : ``{'first', 'last'}``
+            Whether null values should be placed at the beginning
+            or at the end of the result.
+            Note that the position of NaNs is unspecified and may
+            vary based on the implementation.
+
+        Returns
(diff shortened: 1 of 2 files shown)


kkraus14 (approved):

kkraus14:
That performance difference for Polars is quite surprising. Does `df.__getitem__` do a lot of extra bounds checking or something else that could be bypassed in the case where you know the indices being passed to it are generated with guarantees?

rgommers (approved):
This sounds very reasonable to add, LGTM.

It would probably be useful to edit the `sorted_indices` docstrings, since it contains:
```
        If you need to sort the DataFrame, you can simply do::

            df.get_rows(df.sorted_indices(keys))
```
instead, that can now refer to the `sort` method

MarcoGorelli:
> That performance difference for Polars is quite surprising. Does `df.__getitem__` do a lot of extra bounds checking or something else that could be bypassed in the case where you know the indices being passed to it are generated with guarantees?

there are no guarantees in this case

anyway, you can observe the same in numpy:
```python
In [14]: arr = np.random.randint(0, 4, size=100_000_000)

In [15]: %time arr[np.argsort(arr)]
CPU times: user 767 ms, sys: 2.13 s, total: 2.89 s
Wall time: 2.9 s
Out[15]: array([0, 0, 0, ..., 3, 3, 3])

In [16]: %time np.sort(arr)
CPU times: user 253 ms, sys: 187 ms, total: 440 ms
Wall time: 439 ms
Out[16]: array([0, 0, 0, ..., 3, 3, 3])
```

PR #238, opened 2023-08-27
Add dataframe.join

### LINKED ISSUE #237: Support of join operation?
Hey @MarcoGorelli, thanks for this cool project and your talk at EuroScipy 2023!

Do you expect to support joining operations, or is it out of the scope of the dataframe API?

MarcoGorelli:
closes #237

### DIFF
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -897,3 +897,45 @@ class DataFrame:
         understanding that consuming libraries would then use the
         ``array-api-compat`` package to convert it to a Standard-compliant array.
         """
+    
+    def join(
+        self,
+        other: DataFrame,
+        *,
+        how: Literal['left', 'inner', 'outer'],
+        on: str | list[str] | None = None,
+        left_on: str | list[str] | None = None,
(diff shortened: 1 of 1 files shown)


spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -897,3 +897,45 @@ def to_array_object(self, dtype: Any) -> Any:
...
+        """
+        Join with other dataframe.
+
+        Parameters
+        ----------
+        other : DataFrame
+            Dataframe to join with.
+        how : str
+            Kind of join to perform.
+            Must be one of {'left', 'inner', 'outer'}.
```
kkraus14, line 919: I know right joins can be modeled as left joins with the inputs swapped, but any reason not to support `right` as a join type here?

MarcoGorelli: > I know right joins can be modeled as left joins with the inputs swapped

exactly, so if the idea is to make a minimal API, right joins probably aren't needed. have never seem them in practice anyway

kkraus14 (approved):

MarcoGorelli:
cool, thanks for your review!

Issue #212, opened 2023-07-27
Typing `names` argument in `select_columns_by_name`


mroeschke:
Currently `DataFrame.select_columns_by_name` has `names` typed as `names: Sequence[str]`.

I think the intention here is `Sequence` is a non-scalar container of string labels, but `Sequence[str]` also matches a pure `str`.

MarcoGorelli:
hey - this looks correct to me

a `str` is an sequence of `str`. If you have single-letter column names, and pass a string with the column names, I'd expect it to work

and it does:
```python
In [20]: pd.api.interchange.from_dataframe(pd.DataFrame({'a': [1,2,3], 'b': [4,5,6], 'c': [7,8,9]}).__d
    ...: ataframe__().select_columns_by_name('ab'))
Out[20]:
   a  b
0  1  4
1  2  5
2  3  6
```

mroeschke:
Ah okay I forgot about that possibility. Closing then

rgommers:
> `.select_columns_by_name('ab')`

seems pretty ambiguous and bug-prone to me. I'd expect that to give me a single column named `'ab'`, not two columns named `'a'` and `'b'`. I think the intention was for this to be spelled `['a', 'b']`. 

Unfortunately I'm not sure if there's a way to fix this to make it unambiguous while allowing all non-string sequences. And `list[str, ...]` may be too restrictive. So I guess we have to leave it as is either way.

PR #240, opened 2023-08-27
Prefer `str | list[str]` over `Sequence[str]`

MarcoGorelli:
see https://github.com/data-apis/dataframe-api/issues/212

This would make it clearer that you can either pass a single column name, or a list of column names.

`Sequence[str]` would accept `'abcd'` but it's ambiguous whether that means "column 'abcd'" or "columns 'a', 'b', 'c', 'd'", given that `str` is itself a `Sequence[str]`

### DIFF
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -65,13 +65,13 @@ class DataFrame:
         Return number of rows and number of columns.
         """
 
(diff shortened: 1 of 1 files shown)


rgommers (approved):
LGTM, thanks Marco. +1 on being able to write `keys='colname'`

kkraus14:
Are there any concerns about people wanting to be able to pass iterables other than list types? I.E. tuples or anything else?

Issue #224, opened 2023-08-03
How to note that some methods aren't available for lazy dataframes?


MarcoGorelli:
Here are some examples which ~don't work~ raise with polars lazyframes:

- DataFrame.shape (the length isn't, and can't be, known)
- `Column.__len__` (same as above)
- comparisons (e.g. df1 == df2, ~but there may be a way around this~)
- anything column-related (as there's no polars.LazyColumn - but there may be a way around this)

I'd like to think we can work round points ~3 and~ 4. But not 1 and 2, not really sure how that could work

I know we want everything to be independent of the execution engine, but how do we plan for these to work for lazy engines?

betatim:
There was a discussion in the Array API repo a while back that was somewhat similar. It was about things like `__bool__` and `__float__` and how to deal with these for lazy array libraries. https://github.com/data-apis/array-api/issues/642#issuecomment-1622031104 concluded that it was Ok that if a user accessed those a computation was forced.

From the point of view of someone using the dataframe API to write a tool that consumes dataframes (lazy or eager) you don't want to have to distinguish in your code whether a dataframe is lazy or not. It makes me think that if a user wants to know the shape or len, then they want to know the shape/len now. So a lazy dataframe would have to go away and evaluate enough to be able to answer that query. A bit like with `bool(some_array)`, I want to know the answer now, not at some point later.

rgommers:
@MarcoGorelli what does "don't work" mean? Does it force evaluation, raise an Exception, or ....? I think either choice is valid - e.g., Dask will force evaluation when needed, while the comment @betatim linked to had an implementation that needed to be 100% lazy and hence had to raise.

Also, I think `DataFrame.shape()` is not a good example - I don't see a reason why that cannot remain lazy. The reason `bool()`, `int()` et al. couldn't stay lazy is because the Python language itself will not allow returning anything other than actual `bool`/`int`/etc. instances. For `.shape()`, it's perfectly okay to return a lazy object.

MarcoGorelli:
Apologies, by "doesn't work" I meant that it raises an exception

I think I'd be -1 on forcing computation for `.shape` (or anything else). Right now, if you have a `polars.LazyFrame`, then everything is guaranteed to be practically zero-cost up until you hit `.collect()`. If some operations can trigger potentially very expensive operations, then that guarantee is broken

So the alternatives would be:
- raise for `.shape`
- "For .shape(), it's perfectly okay to return a lazy object."

[...]

PR #242, opened 2023-08-27
Remove cross-dataframe comparisons

MarcoGorelli:
Just out of interest, has any here ever needed to sum two entire dataframes? I haven't, and don't think there's much use for these comparison ops when `other` is a dataframe

It also avoids some of the discussion in #224 - if it can't be done, there's no need to document that in some cases it might not be allowed

### DIFF
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -326,7 +326,7 @@ class DataFrame:
         """
         ...
 
-    def __eq__(self, other: DataFrame | Scalar) -> DataFrame:  # type: ignore[override]
+    def __eq__(self, other: Scalar) -> DataFrame:  # type: ignore[override]
         """
         Compare for equality.
 
@@ -334,8 +334,7 @@ class DataFrame:
 
         Parameters
         ----------
-        other : DataFrame or Scalar
-            If DataFrame, must have same length and matching columns.
+        other : Scalar
             "Scalar" here is defined implicitly by what scalar types are allowed
             for the operation by the underling dtypes.
 
@@ -345,7 +344,7 @@ class DataFrame:
         """
         ...
 
-    def __ne__(self, other: DataFrame | Scalar) -> DataFrame:  # type: ignore[override]
+    def __ne__(self, other: Scalar) -> DataFrame:  # type: ignore[override]
         """
         Compare for non-equality.
 
@@ -353,8 +352,7 @@ class DataFrame:
 
         Parameters
         ----------
-        other : DataFrame or Scalar
-            If DataFrame, must have same length and matching columns.
+        other : Scalar
             "Scalar" here is defined implicitly by what scalar types are allowed
             for the operation by the underling dtypes.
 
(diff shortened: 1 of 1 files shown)


rgommers:
I did a quick search only for `df ==` on projects I had a local clone of. A large majority of the hits are comparison with scalars. I did find examples in both scikit-learn [here](https://github.com/scikit-learn/scikit-learn/blob/8357243bf9ced15eec918f9b9d77c1af082ed437/sklearn/tests/test_base.py#L376) and pandas [here](https://github.com/pandas-dev/pandas/blob/ca429943c70cee2077941f052281b92afc0c2054/doc/source/whatsnew/v0.24.0.rst#dataframe-comparison-operations-broadcasting-changes). 

I guess the main questions I have are: 
- doesn't the same thing apply to the corresponding methods of columns?
- if a user does need a comparison or element-wise math operation, then how do they do it if we merge this PR?

MarcoGorelli:
Thanks for taking a look

The pandas example is just in the release notes
And the scikit-learn one is just in a test, where they should probably be using `pd.testing.assert_frame_equal`
I'd be OK with adding testing functionality to the standard, performance is far less of a concern there anyway

Do we have any examples of downstream libraries using this in their core code?

> if a user does need a comparison or element-wise math operation, then how do they do it if we merge this PR?

They do a join beforehand. The reason to not do a join under-the-hood for them is that like this, they only do the join once, and then can do all the comparisons maths they want

> doesn't the same thing apply to the corresponding methods of columns?

If the columns come from the same dataframe, then no, it'll all work fine
If the columns come from different dataframes, then yes, the dataframes would need joining beforehand

Try expressing, using just sql syntax, the sum between `col1` and `col2`. If they're from the same table:
```sql
select col1 + col2 as 'col1_col2_sum'
from df
```
If they're not, then a join is required

kkraus14:
I'm generally a +1 removing `DataFrame.__eq__` and other binary operators against other DataFrames, but given this constraint:

> If the columns come from the same dataframe, then no, it'll all work fine
> If the columns come from different dataframes, then yes, the dataframes would need joining beforehand

I think we need to work through how we handle Columns more generally, because our current design doesn't have this kind of information.

MarcoGorelli:
Agree that we need to work through how we handle columns more generally, and that needs doing independently of this PR

We OK to start with this one?

rgommers:
One thing I came to realize is that there are two types of dataframe usage:

1. The main one: SQL-like, lots of high-level operations like merge/join/select/pivot/apply,
2. A historically common (but waning in populariy) one: "2-D arrays with axis labels and masks"

Element-wise operations like `==` and `+` are for (2) rather than (1). This is the kind of usage scikit-learn has - essentially they're treating dataframes like arrays with labels, and internally they still use arrays and then re-attach labels before returning a dataframe.

@MarcoGorelli and I discussed this a bit yesterday, and concluded that this would be okay to drop as a mode of operation (and hence, dropping `==` and other element-wise operations as done in this PR). That use case isn't core to today's dataframe libraries, and scikit-learn & co will not actually need this since it's still the norm (and probably will always be more efficient) to unpack the data and then use an array library for any heavy lifting.

Probably this PR can be merged after tomorrow's meeting.

rgommers (approved):
This got more thumbs-up's in the meeting yesterday, and it seems like in general we'd like to revisit/remove the APIs that have now been identified as being problematic for lazy implementations. So in it goes. Thanks @MarcoGorelli and @kkraus14!

PR #245, opened 2023-08-30
rename `get_columns_by_name` to `select`

MarcoGorelli:
"get_columns_by_name" has been annoying me for a while. Someone has even commented that there's no need to note that this API is developer-focused, as with such an anti-ergonomic name, no user would ever consider using it anyway 😄 
Then, the following question from a user has prompted me to finally suggest renaming it:

image

I'm suggesting `select` because it's already the standardised name:
- in SQL, you select columns with `select`
- pyspark has `DataFrame.select`
- polars has `DataFrame.select`

### DIFF
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -109,7 +109,7 @@ class DataFrame:
         """
         ...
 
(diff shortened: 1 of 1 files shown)


MarcoGorelli:
this was verbally approved in the last call

if there's no objections, will merge later this week

PR #246, opened 2023-08-30
Rename `get_rows_by_mask` to `filter`

MarcoGorelli:
Similar to #245

In for a penny, in for a pound - if we're gonna make breaking changes whilst we're still in "beta", I suggest we go all the way there

Areg had said that it would be nice to get to the point of making this user-friendly, and names like `get_columns_by_name` and `get_rows_by_mask` aren't really going to cut it

why `filter`? Because it's already there in:
- pyspark
- polars

### DIFF
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -104,7 +104,7 @@ class Column(Generic[DType]):
         ...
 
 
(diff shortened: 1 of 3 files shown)


MarcoGorelli:
this was verbally approved in the last call

if there's no objections, will merge later this week

PR #248, opened 2023-08-31
Add dataframe.schema

MarcoGorelli:
Looks like this is missing


rgommers (approved):
LGTM, and is quite simple/straightforward. I suggest we merge this within the next couple of days unless there are new comments with a concern.

PR #254, opened 2023-09-06
Rename `get_column_names` to `column_names` and make property

MarcoGorelli:
Similar to #245 and #246, trying to improve ergonomics whilst we're still "beta"

I think we've agreed that although the number of rows (and their ordering) isn't always known, we do always know the number of columns (if we didn't, then `get_column_names` would already be broken), and that this shouldn't really involve any compute


MarcoGorelli:
there weren't any objections last time, and `schema` is already a property, and nobody's commented in 3 weeks, so I'll go ahead and merge

PR #256, opened 2023-09-07
Note column names uniqueness constraint on join

MarcoGorelli:


spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -932,6 +932,10 @@ def join(
         """
         Join with other dataframe.
 
+        Other than the joining column name(s), no column name is allowed to appear in
+        both `self` and `other`. Rename columns before calling `join` if necessary
+        using :meth:`rename_columns`.
```
MarcoGorelli, line 937: todo: note `raises`

MarcoGorelli:
i've noted what should happen with regards to raising - any objections

MarcoGorelli:
this was discussed last time, and I addressed the one comment, so let's get it in

PR #259, opened 2023-09-13
Rename groupby to group_by

MarcoGorelli:
This is what the ecosystem generally does:

- SQL has `GROUP BY`
- pyspark has `groupBy` (note that the `B` is capitalised, as it's the start of a new word)
- dplyr has `group_by`
- polars has `group_by`
- it's two words

I guess the only reason it's currently `groupby` is because that's what pandas did?

But that API is now so far away from pandas anyway, so I suggest we follow the broader ecosystem


MarcoGorelli:
verbally approved in the last call, and nobody's objected - merging then, but if anyone objects I can revert

PR #263, opened 2023-09-27
Return namespace

MarcoGorelli:
having to keep typing `__dataframe_namespace__` is quite inconvenient, shall we just return it right away (as is done in the array api)?

### DIFF
--- a/protocol/purpose_and_scope.md
+++ b/protocol/purpose_and_scope.md
@@ -50,21 +50,21 @@ aims to develop in the future. When that full API standard is implemented by
 dataframe libraries, the example above can change to:
 
 ```python
-def get_compliant_df(df):
+def get_compliant_df_and_namespace(df) -> tuple[DataFrame, Any]:
     """Utility function to support programming against a dataframe API"""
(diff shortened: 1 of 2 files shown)


rgommers:
In my experience it doesn't really matter, when you use this it's anyway a little unusual to call a dunder method directly. I think what you end up doing as writing a single convenience function that does more things (input validation, processing multiple dataframes at once, etc.). E.g.: https://github.com/scipy/scipy/blob/45e875d819c24a3adaa187a1d7050c164d3e937d/scipy/_lib/_array_api.py#L74-L108

This would also necessitate updates to these methods in Pandas/Polars/etc, right? Is it all really worth doing?

MarcoGorelli:
> This would also necessitate updates to these methods in Pandas/Polars/etc, right?

they just defer to `dataframe-api-compat`, and are typed to return `Any` anyway, so I can just handle it in `dataframe-api-compat` directly

(if I'd had to make changes upstream then I would definitely not have suggested this 😄)

MarcoGorelli:
rejected on the call (this isn't actually what the array api does)

PR #264, opened 2023-09-27
remove some unnecessary api_version args

MarcoGorelli:
These are unnessary, as the namespace will already have been instantiated to follow a given api version

e.g.

```python
df, xp = df.__dataframe_consortium_standard__(api_version='2023.09-beta')
# xp is already the 2023.09-beta API version
# so xp.dataframe_from_dict returns a dataframe compliant with the 2023.09-beta version
```


MarcoGorelli:
approved on today's call

PR #265, opened 2023-09-27
remove unused typevars, make dtype type alias

MarcoGorelli:
Summary:
- make `DType` a type alias
- remove several unused TypeVars (we'd agreed to disagree on removing them in https://github.com/data-apis/dataframe-api/pull/198/files, but some still remain)

### DIFF
--- a/spec/API_specification/dataframe_api/__init__.py
+++ b/spec/API_specification/dataframe_api/__init__.py
@@ -3,12 +3,15 @@ Function stubs and API documentation for the DataFrame API standard.
 """
 from __future__ import annotations
 
-from typing import Mapping, Sequence, Any
+from typing import Mapping, Sequence, Any, TYPE_CHECKING
 
 from .column_object import *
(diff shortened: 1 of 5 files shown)


rgommers (commented):
Thanks @MarcoGorelli. This LGTM overall. In particular, going from `dtype: Any` to `dtype: DType` seems nice.

My one question is about removing support for parametrizing `Column` by its dtype (`Column[Bool]` et al). Do you see that as an improvement, or a (perhaps temporary) regression?

MarcoGorelli:
a teeny tiny regression perhaps, but I think the `DType` type alias would be more useful to devs, so I'd happily prioritise that one

in any case, if we did go ahead with #249 , then individual columns would become less relevant (and would probaby only be a last step before converting to ndarray anyway, or an initial step during construction)

rgommers (approved):
Okay, SGTM! This affects only static typing and html docs, rather than an API change. So let me hit the green button here.

PR #266, opened 2023-09-27
add namespace protocol

### LINKED ISSUE #258: Bad developer experience - `df.__dataframe_namespace__()` is untyped
The developer experience of using the api could be improved (to put it mildly) - I'll give a demo

If I stick to dataframe methods, then it's all typed and so I get nice suggestions from my IDE:

image

But if I try getting the namespace out, I don't get any suggestions, because `__dataframe_namespace__()` is typed to return `Any`

image

Before I make suggestions, can we agree that this is a bad developer experience and that it needs fixing?

MarcoGorelli:
closes #258

### DIFF
--- a/spec/API_specification/dataframe_api/__init__.py
+++ b/spec/API_specification/dataframe_api/__init__.py
@@ -3,12 +3,15 @@ Function stubs and API documentation for the DataFrame API standard.
 """
 from __future__ import annotations
 
-from typing import Mapping, Sequence, Any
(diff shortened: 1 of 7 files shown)


rgommers:
This needs a rebase now

MarcoGorelli:
taking this forwards as part of https://github.com/data-apis/dataframe-api/pull/267 (it's practically impossible to know if I've done this correctly without some concrete examples to type-check)

PR #267, opened 2023-09-27
Add examples

MarcoGorelli:
needs rebasing onto https://github.com/data-apis/dataframe-api/pull/266

---

It's going to be virtually impossible to keep the example in the docs up-to-date, so I suggest we make a separate `examples` folder, which we can type check in CI and make sure it keeps working


rgommers:
+1 in the call today about having examples, in a separate folder so they're easy to keep up to date in CI and don't get mixed with the language of the actual standard.

This PR needs a rebase.

MarcoGorelli:
thanks - approved during call

there may be some minor issues but I guess they'll become apparent as we add more examples

PR #269, opened 2023-09-28
update/insert columns -> assign

### LINKED ISSUE #235: `set_column_by_name`/ `set_column`?
Thanks for the work that you're doing here!

I would like to request the addition of `DataFrame.set_item_by_name` to the API. 

From the demonstration at euroscipy last week, the ugliest part of the usage was adding columns by:

```python
for col in cols:
    df.insert(len(df.get_column_names()), col)
```

This seems like an obvious omission, but I haven't found any documentation saying why it hasn't been included. Seems like it was discussed in a call (https://github.com/data-apis/dataframe-api/pull/138#issue-1670699315)?

MarcoGorelli:
closes #235 

I think I made a mistake in splitting "set_column" into "insert_column" and "update_column":
1. first, because having tried using the Standard to write code a while now, it's really tedious to have to distinguish them
2. because I don't think any existing dataframe library has this distinction
3. a users has brought this up too (#235)

In keeping with the goal of making the standard feel familiar, let's look at what existing dataframe libraries do:
- pandas and its clones: either `__setitem__` (which the consortium doesn't want) or `DataFrame.assign`
- pyspark / polars: `DataFrame.with_columns`
- ibis: `DataFrame.mutate`
- datatable: `update / extend`
 
I don't really mind which one we go with. I'm suggesting `with_columns` here as I think it's probably the most explicit

### DIFF
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -180,53 +180,27 @@ class DataFrame:
         """
         ...
 
-    def insert_column(self, column: Column[Any]) -> DataFrame:
+    def with_columns(self, columns: Column[Any] | Sequence[Column[Any]], /) -> DataFrame:
         """
-        Insert column into DataFrame at rightmost location.
+        Insert new column(s), or update values in existing ones.
 
-        The column's name will be used as the label in the resulting dataframe.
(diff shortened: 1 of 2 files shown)


MarcoGorelli:
approved in today's call (but consensus on `assign` instead of `with_columns`)

PR #275, opened 2023-10-05
Add basic datetime functionality

### LINKED ISSUE #260: Datetime functionality
I like how pandas/polars have a `.dt` namespace for datetime functionality, I'd suggest having that too. What do we put in it?

Looking at skrub, here's some datetime functionality we should add.

- year
- month
- day
- hour
- minute
- second
- millisecond
- microsecond
- nanosecond
- iso_weekday (monday = 1, sunday = 7)
- timestamp (number of seconds since 1970-01-01 UTC)

These should all be fairly trivial. They also use `floor`, but I think we should keep that one out initially, as it's really non-trivial in the timezone-aware case when there's DST. Pretty sure there's a way around it for what they're doing anyway

There's some inconsistency in the definitions here:
- pandas: `microsecond` returns the number of microseconds since the last second, but `'nanosecond'` returns the number of nanoseconds since the last microsecond. there is no `millisecond`
- polars (and chrono): `nanosecond` returns the number of nanoseconds since the last second. similarly for `millisecond` and `microsecond`

I'd suggest we only include `microsecond` as part of the Standard, which they all agree on

Also, I'd suggest making all these functions rather than properties. I think pandas is really misleading here:
```python
In [71]: ts = pd.date_range('1900-01-01', '2100-01-01', freq='1min')

In [72]: %time ts.nanosecond  # definitely not "free"
CPU times: user 2.1 s, sys: 26 ms, total: 2.13 s
Wall time: 2.13 s
Out[72]:
Index([0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
       ...
       0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
      dtype='int32', length=105190561)
```

MarcoGorelli:
closes #260


spec/API_specification/dataframe_api/column_object.py
```diff
@@ -763,3 +765,106 @@ def rename(self, name: str) -> Column:
...
+        """
+
+    def microsecond(self) -> Column:
+        """
+        Return number of microseconds since last second, for each element.
+
+        For example, return 123456 for 1981-01-02T12:34:56.123456.
+        """
+    
+    def floor(self, frequency: timedelta) -> Column:
```
kkraus14, line 830: I'm -1 on ducktyping `datetime.timedelta` objects as our scalar for timedeltas. It does a lot of weird rounding, i.e.
```
from datetime import timedelta

print(timedelta(microseconds=1.1111).microseconds)  # 1
```

More generally, I don't think it's friendly to ducktype and doesn't really align with how many dataframe libraries handle timedeltas.

MarcoGorelli: what would you suggest instead?

kkraus14: I think we could keep the interface of `timedelta` but would need to possibly redefine some of the error cases, behaviors, and semantics?

Some examples where I think there's problems:

* It can't handle large integer values for milliseconds and other units which we'd presumably want to support?
```
from datetime import timedelta

int64_max = 9223372036854775807

timedelta(milliseconds=int64_max)  # Throws OverflowError: Python int too large to convert to C int
```

* Construction from floats is allowed and introduces rounding behavior that everyone would need to match
```
from datetime import timedelta

timedelta(microseconds=0.5).microseconds  # 0
timedelta(microseconds=1.5).microseconds  # 2
```

* We may eventually want to extend support for units that it doesn't support like nanoseconds as many libraries support it (Pandas, PyArrow, cuDF, etc.).
* Would there be any issues representing a `NULL` scalar of timedelta type? This would be distinctly different than our `null` object/type since it could be returned by something like a `get_value` or aggregation function.

I'm sure there's more, but just from an initial pass.

MarcoGorelli: How about just strings? Like `frequency="1minutes"` or `frequency="1days30minutes"`

MarcoGorelli:
I've removed `floor`, we can discuss that one later, let's start with the simple ones

rgommers (commented):
Thanks Marco. This all looks pretty logical. It'd be useful to add to the spec explicitly that the returned columns here all have `int64` (I think?) dtype.

MarcoGorelli:
Thanks

Not necessarily, some of these could be unsigned

spec/API_specification/dataframe_api/column_object.py
MarcoGorelli: taking this forwards in https://github.com/data-apis/dataframe-api/pull/277

PR #276, opened 2023-10-07
Replace dataframe_from_dict with dataframe_from_columns

MarcoGorelli:
`Column`s have names, so `dataframe_from_dict` isn't really necessary

drive-by: removing `api_version` from the functions in the `Namespace` protocol, which I should've done in https://github.com/data-apis/dataframe-api/pull/264


kkraus14 (approved):

MarcoGorelli:
thanks for your review!

PR #278, opened 2023-10-08
Make dataframe-api installable

MarcoGorelli:
Can we make the dataframe-api installable please? This would make development of the implementation much more pleasant


rgommers (approved):
This seems very reasonable, and I can't think of any downside or constraint that this introduces, so I think it can be merged without needing wider discussion. In it goes - thanks Marco.

PR #284, opened 2023-10-17
Allow for positional arguments instead of str | list[str]

MarcoGorelli:
I'd like to suggest passing multiple columns positionally, rather than as `str | list[str]`.

First, this is more ergonomic to write, as it saves an extra pair of brackets (and an extra level
of indentation...).

Second, because this is actually already supported in both pandas and Polars:
```python
# pandas

In [2]: df.assign(
   ...:     d=lambda x: x['a']*2,
   ...:     e=lambda x: x['b'] - x['a']
   ...: )
Out[2]:
   a  b  c  d  e
0  1  4  4  2  3
1  2  5  2  4  3
2  3  6  1  6  3

# Polars

In [1]: df.with_columns(
   ...:     d=pl.col('a')*2,
   ...:     e=pl.col('b')-pl.col('a')
   ...: )
Out[1]:
shape: (3, 5)
┌─────┬─────┬─────┬─────┬─────┐
│ a   ┆ b   ┆ c   ┆ d   ┆ e   │
│ --- ┆ --- ┆ --- ┆ --- ┆ --- │
│ i64 ┆ i64 ┆ i64 ┆ i64 ┆ i64 │
╞═════╪═════╪═════╪═════╪═════╡
│ 1   ┆ 4   ┆ 4   ┆ 2   ┆ 3   │
│ 2   ┆ 5   ┆ 2   ┆ 4   ┆ 3   │
│ 3   ┆ 6   ┆ 1   ┆ 6   ┆ 3   │
└─────┴─────┴─────┴─────┴─────┘
```

Third, because it doesn't restrict multiple arguments to being `list` - now, as long as they're an iterable,
they can be passed.

(Reminder: the reason we can't have `str | Iterable[str]` is that in Python `str` is itself an `Iterable[str]`,
so it's ambiguous whether `columns='abc'` means "just column 'abc'" or "columns 'a', 'b', and 'c'".)

### DIFF
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -794,7 +794,7 @@ class DataFrame:
         """
         ...
 
-    def unique_indices(self, *keys: str, *, skip_nulls: bool = True) -> Column:
+    def unique_indices(self, *keys: str, skip_nulls: bool = True) -> Column:
         """
         Return indices corresponding to unique values across selected columns.
 
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -64,13 +64,13 @@ class DataFrame:
         Return number of rows and number of columns.
         """
 
-    def group_by(self, keys: str | list[str], /) -> GroupBy:
+    def group_by(self, *keys: str) -> GroupBy:
         """
         Group the DataFrame by the given columns.
 
         Parameters
         ----------
-        keys : str | list[str]
+        *keys : str
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


kkraus14:
I completely agree with the intention of this PR to always treat the column keys as something "Iterable-like". Is there ever a situation where someone wants to pass something "Column-like" to any of these parameters where we then can't iterate (and implementations would maybe want to take advantage of a non-Python fastpath)?

MarcoGorelli:
thanks

not sure what you're asking sorry, could you give an example pelase?

kkraus14:
> thanks
> 
> not sure what you're asking sorry, could you give an example please?

I was thinking that there might be situations where you'd want to feed a `Column[str]` into any of these APIs, but we're consistently using Python strings for column names where this just makes sense. Approving.

kkraus14 (approved):

MarcoGorelli:
thanks - moving forwards then as this is relatively minor

PR #285, opened 2023-10-17
Add TPC-H query 5 to examples

MarcoGorelli:
needs rebasing onto #282 when that's in

### DIFF
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -6,7 +6,7 @@ from typing import Any, Literal, Mapping, Sequence, Union, TYPE_CHECKING, NoRetu
 if TYPE_CHECKING:
     from .column_object import Column
     from .groupby_object import GroupBy
-    from .typing import NullType, Scalar, Namespace, DType
+    from .typing import NullType, Scalar, Namespace, DType, SupportsDataFrameAPI
 
 
 __all__ = ["DataFrame"]
@@ -51,7 +51,7 @@ class DataFrame:
         """
 
     @property
-    def dataframe(self) -> object:
+    def dataframe(self) -> SupportsDataFrameAPI:
         """
         Return underlying (not-necessarily-Standard-compliant) DataFrame.
 
--- a/spec/API_specification/examples/tpch/q5.py
+++ b/spec/API_specification/examples/tpch/q5.py
@@ -1,0 +1,57 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from dataframe_api.typing import DataFrame, SupportsDataFrameAPI
+
+
+def query(
+    customer_raw: SupportsDataFrameAPI,
+    orders_raw: SupportsDataFrameAPI,
+    lineitem_raw: SupportsDataFrameAPI,
+    supplier_raw: SupportsDataFrameAPI,
+    nation_raw: SupportsDataFrameAPI,
+    region_raw: SupportsDataFrameAPI,
+) -> Any:
+    customer = customer_raw.__dataframe_consortium_standard__()
+    orders = orders_raw.__dataframe_consortium_standard__()
+    lineitem = lineitem_raw.__dataframe_consortium_standard__()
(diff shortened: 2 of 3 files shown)


spec/API_specification/examples/tpch/q5.py
```diff
@@ -0,0 +1,58 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from dataframe_api.typing import DataFrame, SupportsDataFrameAPI
+
+
+def query(
```
kkraus14, line 9: Would be good to add the SQL query as a docstring or comment just for reference:
```sql
select
  n_name,
  sum(l_extendedprice * (1 - l_discount)) as revenue
from
  customer,
  orders,
  lineitem,
  supplier,
  nation,
  region
where
  c_custkey = o_custkey
  and l_orderkey = o_orderkey
  and l_suppkey = s_suppkey
  and c_nationkey = s_nationkey
  and s_nationkey = n_nationkey
  and n_regionkey = r_regionkey
  and r_name = 'ASIA'
  and o_orderdate >= '1994-01-01'
  and o_orderdate < '1995-01-01'
group by
  n_name
order by
  revenue desc
```

spec/API_specification/examples/tpch/q5.py
```diff
@@ -0,0 +1,58 @@
...
+    customer_raw: SupportsDataFrameAPI,
+    orders_raw: SupportsDataFrameAPI,
+    lineitem_raw: SupportsDataFrameAPI,
+    supplier_raw: SupportsDataFrameAPI,
+    nation_raw: SupportsDataFrameAPI,
+    region_raw: SupportsDataFrameAPI,
+) -> Any:
```
kkraus14, line 16: Could we have this return a `SupportDataFrameAPI` instead of calling `.dataframe` at the end?

spec/API_specification/examples/tpch/q5.py
```diff
@@ -0,0 +1,58 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from dataframe_api.typing import DataFrame, SupportsDataFrameAPI
```
kkraus14, line 6: The `DataFrame` type doesn't seem to be used here?

```suggestion
    from dataframe_api.typing import SupportsDataFrameAPI
```

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -51,7 +51,7 @@ def __dataframe_namespace__(self) -> Namespace:
         """
 
     @property
-    def dataframe(self) -> object:
+    def dataframe(self) -> SupportsDataFrameAPI:
```
kkraus14, line 54: I don't think this is correct, since we defined this API specifically as returning a not necessarily standard compliant dataframe?

spec/API_specification/examples/tpch/q5.py
```diff
@@ -0,0 +1,74 @@
...
+
+    new_column = (
+        result.get_column_by_name("l_extendedprice")
+        * (1 - result.get_column_by_name("l_discount"))
+    ).rename("revenue")
+    result = result.assign(new_column)
+    result = result.select(["revenue", "n_name"])
+    result = result.group_by("n_name").sum()
+
+    return result.dataframe
```
kkraus14, line 74: Why return `result.dataframe` instead of just `result` here?

spec/API_specification/dataframe_api/dataframe_object.py
MarcoGorelli: `SupportsDataFrameAPI` just means that it has a `__dataframe_consortium_standard__` attribute, which the return value would have

the idea: pandas dataframe in, pandas dataframe out. fancydataframe in, fancydataframe out. etc... . the user of this function need not know that the dataframe api was used within the body of the function

spec/API_specification/examples/tpch/q5.py
MarcoGorelli: so that the input class is preserved  https://github.com/data-apis/dataframe-api/pull/285#discussion_r1367781075

kkraus14 (approved):

PR #286, opened 2023-10-17
Add GroupBy.aggregate (and tpch-1 query to examples)

### LINKED ISSUE #274: GroupBy.aggregate
I think the current group_by design is too limited. For example, trying to implement tpc-h queries, how are we supposed to write
```sql
SELECT l_returnflag,
       l_linestatus,
       Sum(l_extendedprice * ( 1 - l_discount )) AS sum_disc_price,
       Avg(l_quantity)                           AS avg_qty,
       Count(*)                                  AS count_order
FROM   lineitem
GROUP  BY l_returnflag,
          l_linestatus  
```
?

There's two problems with the current API:
1. we can't mix and match `sum` and `avg` within the same groupby
2. we can't transform columns within the group by aggregation. For example, tpc-h would require `l_extendedprice*(1-l_discount)` to be written inside the group by, rather than outside it

There's a couple of solutions I can think of

## 1. Expressions

Once again, the expressions API may help us here (https://github.com/data-apis/dataframe-api/pull/249)

We could introduce `GroupBy.aggregate`, which you can pass an expression to and which must reduce itself to a 1-row column per group

So, for example, the above could be expressed with:
```python
lineitem: DataFrame
lineitem.group_by('l_return_flag', 'l_linestatus').aggregate(
    (col('l_extendedprice')*(1-col('discount'))).sum().rename('sum_disc_price'),
    col('l_quantity').mean().rename('avg_qty'),
    col('l_return_flag').count().rename('count_order'),
)
```

In the pandas implementation:
- `col('l_quantity').mean().rename('avg_qty')` could get mapped to `avg_qty=pd.NamedAgg(column="l_quantity", aggfunc="mean")`
- `(col('l_extendedprice')*(1-col('discount'))).sum().rename('sum_disc_price')` would probably require a custom lambda function

This would be maximally flexible, but it would mean that people would write code which could be inefficient in the pandas implementation

## 2. (output_column_name, input_column_name, aggregation_name, **kwargs) tuples

Alternatively, `.aggregate` could take a tuple of (output_column_name, input_column_name, aggregation_name), and in the case above then `(col('l_extendedprice')*(1-col('discount')))` would have had to be computed before entering the `group_by`. E.g.:
```python
lineitem: DataFrame
lineitem = lineitem.assign(
    (col('l_extendedprice')*(1-col('discount'))).rename('disc_price')
)
lineitem.group_by('l_return_flag', 'l_linestatus').aggregate(
    ('sum_disc_price', 'disc_price', 'sum', {}),
    ('avg_qty', 'l_quantity', 'mean', {}),
    ('count_order', 'l_return_flag', 'count', {}),
)
```
Maybe we could also define our own dataclass to make the above more readable

Technically this might not be totally in accordance to tpc-h rules on not reordering, but it would avoid requiring lambda functions for the pandas implementation

---

Which solution is best?
- if we're going for "how should the API of a new dataframe library look like?", then I prefer option 1
- if we're going for "a minimal API which all major dataframe libraries today can agree on" then I prefer option 2

All things considered, I might be leaning more towards option 2

MarcoGorelli:
closes #274 

the gist of the PR is that it lets you write
```python
    result = (
        lineitem.filter(mask)
        .group_by(["l_returnflag", "l_linestatus"])
        .aggregate(
            namespace.Aggregation.sum("l_quantity").rename("sum_qty"),
            namespace.Aggregation.sum("l_extendedprice").rename("sum_base_price"),
            namespace.Aggregation.sum("l_disc_price").rename("sum_disc_price"),
            namespace.Aggregation.sum("change").rename("sum_charge"),
            namespace.Aggregation.mean("l_quantity").rename("avg_qty"),
            namespace.Aggregation.mean("l_discount").rename("avg_disc"),
            namespace.Aggregation.size().rename("count_order"),
        )
        .sort(["l_returnflag", "l_linestatus"])
    )
```

### DIFF
--- a/spec/API_specification/dataframe_api/groupby_object.py
+++ b/spec/API_specification/dataframe_api/groupby_object.py
@@ -75,7 +75,11 @@ class GroupBy:
 
 class Aggregation(Protocol):
     def rename(self, name: str) -> Aggregation:
-        """Assign given name to output of aggregation. """
+        """
+        Assign given name to output of aggregation.
+
+        If not called, the column's name will be used as the output name.
+        """
         ...
 
     @classmethod
--- a/spec/API_specification/dataframe_api/groupby_object.py
+++ b/spec/API_specification/dataframe_api/groupby_object.py
@@ -1,12 +1,15 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Protocol
 
 if TYPE_CHECKING:
     from .dataframe_object import DataFrame
 
 
-__all__ = ['GroupBy']
+__all__ = [
+    "Aggregation",
+    "GroupBy",
+]
 
 
 class GroupBy:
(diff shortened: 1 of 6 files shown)


spec/API_specification/examples/tpch/q1.py
```diff
@@ -0,0 +1,39 @@
...
+                lineitem.get_column_by_name("l_extended_price")
+                * (1 - lineitem.get_column_by_name("l_discount"))
+            ).rename("l_disc_price"),
+            (
+                lineitem.get_column_by_name("l_extended_price")
+                * (1 - lineitem.get_column_by_name("l_discount"))
+                * (1 + lineitem.get_column_by_name("l_tax"))
+            ).rename("l_charge"),
+        ]
+    )
```
MarcoGorelli, line 24: this syntax though...

MarcoGorelli: Hi @shwina - are you ok with this syntax?

Personally I think it's worse than what's in any existing dataframe library, and I can't imagine any user ever wanting to write code like this

 but maybe it's just me

shwina: Please correct me if I'm wrong, but I thought the goal of the standard right now is to provide an API for library developers (not end users). This is why we have been comfortable sacrificing syntactic crispness or an expressive API in favor of being the "lowest common denominator" that all libraries can implement.

I think this necessarily means the API isn't quite as nice to work with for the end-user.

For example, changing `get_column_by_name` to just `[ ]` in the code above would be a massive boost in readability, but we explicitly decided against it because (IIRC) we wanted library authors to have the freedom to decide what `[ ]` should mean for their library

shwina: That being said, I agree with you 100% that this looks a mess. It's a question whether library developers are going to be OK with dealing with a messy API to get cross-library compatibility in return...

MarcoGorelli: >  I agree with you 100% that this looks a mess

Well I'm glad we could find some common ground :smile:

Let's discuss more next week - I'm genuinely interested in finding a solution that works for everybody

My current prediction is that, unless the standard drastically improves, that libraries will just support pandas and Polars and ignore the standard completely

The end result for cudf will be that you'll be no better off than you are now

MarcoGorelli: As I was saying...(emphasis mine)

> I'm pretty upset about having to use df.get_column_by_name("a") instead of a simpler df["a"] or col("a"). This will obfuscate our code and impair readability, and therefore **we may consider keeping our duplicate logic**

https://github.com/data-apis/dataframe-api/issues/287

shwina: That's fair. We should shorten the name.

kkraus14: Being addressed in #290

spec/API_specification/dataframe_api/groupby_object.py
```diff
@@ -51,3 +54,71 @@ def var(self, *, correction: int | float = 1, skip_nulls: bool = True) -> DataFr
...
+        ...     namespace.Aggregation.size().rename('count_order'),
+        ... )
+        """
+        ...
+
+class Aggregation(Protocol):
+    def rename(self, name: str) -> Aggregation:
+        """Assign given name to output of aggregation. """
+        ...
+
```
kkraus14, line 80: What is the behavior is someone doesn't call `rename`? Do we have a standard naming scheme where it reuses the column name or something like that?

kkraus14 (approved):
LGTM other than the move to `df.col`

MarcoGorelli:
thanks for your review

we can rename to `col` if/when the other PR is in

PR #290, opened 2023-10-23
rename get_column_by_name to col

### LINKED ISSUE #287: Feature request: `get_column_by_name` impairs readability
Hey dataframe-api team!

First off, thank you for this effort. I'm a developer of [skrub](https://github.com/skrub-data/skrub), a data-wrangling package. We aim to support many dataframes, starting with Pandas and Polars, and currently have duplicate logic.

We think the dataframe-API is the way for us (see [this discussion](https://github.com/skrub-data/skrub/discussions/719)) and hope to make the switch soon.

I followed the discussion of https://github.com/data-apis/dataframe-api/pull/249 and I'm pretty upset about having to use `df.get_column_by_name("a")` instead of a simpler `df["a"]` or `col("a")`. This will obfuscate our code and impair readability, and therefore we may consider keeping our duplicate logic.

To keep an acceptable UX, could we consider having at least a shorter alias to `get_column_by_name`?

MarcoGorelli:
closes #287

### DIFF
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -27,6 +27,6 @@ jobs:
             ~\AppData\Local\pip\Cache
           key: ${{ runner.os }}-build-${{ matrix.python-version }}
       - name: install-reqs
-        run: python -m pip install --upgrade mypy==1.4.0
+        run: python -m pip install --upgrade mypy==1.4.0 typing-extensions
       - name: run mypy
         run: cd spec/API_specification && mypy dataframe_api && mypy examples
--- a/spec/API_specification/.mypy.ini
+++ b/spec/API_specification/.mypy.ini
@@ -1,5 +1,2 @@
 [mypy]
 strict=True
-
-[mypy-dataframe_api.*]
-disable_error_code=empty-body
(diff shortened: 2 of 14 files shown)


kkraus14 (approved):

shwina (approved):
I'm +1 to this, but would prefer if we brought this up in the next meeting before merging. Thanks!

MarcoGorelli:
sure, will wait, thanks

rgommers (approved):
For context, I think we inherited the `get_column_by_name` method naming from the interchange protocol. For that protocol, usage is kind of a "one shot" thing, where you may use a few methods when doing a conversion, but it's not like you write a whole library using that API. There is no real reason why we have to keep names that are in the interchange protocol, and `get_column_by_name` is particularly cumbersome, so +1 to renaming. Any reasonable and ergonomic name works here (`.col`, `.get_col`, etc.), so +1 from me.

The one thing to keep in mind for future renames is probably that we should avoid having method names that are also part of the protocol if the semantics are different. But that doesn't seem too likely.

MarcoGorelli:
approved in today's call too - thanks all, let's get this in and make things happen

PR #292, opened 2023-10-23
use other: Self instead of other: Column

MarcoGorelli:
We're not saying that
```python
s1: PandasColumn
s2: FancyColumn
s1 + s2
```
is supported, so let's use `Self`

Else implementations will get
```console
t.py:18: error: Argument 1 of "add" is incompatible with supertype "Cat"; supertype defines the argument type as "Cat"  [override]
t.py:18: note: This violates the Liskov substitution principle
t.py:18: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
t.py:21: error: Argument 1 of "sub" is incompatible with supertype "Cat"; supertype defines the argument type as "Cat"  [override]
t.py:21: note: This violates the Liskov substitution principle
t.py:21: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
Found 2 errors in 1 file (checked 1 source file)
```

### DIFF
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -27,6 +27,6 @@ jobs:
             ~\AppData\Local\pip\Cache
           key: ${{ runner.os }}-build-${{ matrix.python-version }}
       - name: install-reqs
-        run: python -m pip install --upgrade mypy==1.4.0
+        run: python -m pip install --upgrade mypy==1.4.0 typing-extensions
       - name: run mypy
         run: cd spec/API_specification && mypy dataframe_api && mypy examples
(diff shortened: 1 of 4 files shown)


kkraus14:
What happens if a library wants to allow using other libraries columns as well as their own?

I.E. cudf may want to allow any CPU or other NVIDIA GPU library as input, assuming we eventually tie the interchange protocol and API together similarly to what was done in the array API.

MarcoGorelli:
that would be fair enough, but I'm not sure it should be part of the standard?

typing `other: Column` in theory means "any object which implements the `Column` protocol can be passed here", which even in the cudf case wouldn't be true

kkraus14 (approved):

MarcoGorelli:
thanks for your review!

PR #293, opened 2023-10-23
rename to_array_object to to_array, remove `dtype` argument in Column case

MarcoGorelli:
Couple of things here:

- `to_array_object` is quite long. The docstring already describes what exactly the method does. Could it just be `to_array` instead?
- In the Column case, can we remove the `dtype` argument? Much simpler to just convert to the respective dtype, and "force" users to fill nulls before hand

### DIFF
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -732,27 +732,26 @@ class Column:
         """
         ...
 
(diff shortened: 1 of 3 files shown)


kkraus14 (approved):

MarcoGorelli:
thanks for your review!

rgommers:
A belated +1 from me - these changes look useful.

Issue #281, opened 2023-10-17
Request for more examples


MarcoGorelli:
For the dataframe Api we currently have a folder with some examples https://github.com/data-apis/dataframe-api/tree/main/spec/API_specification/examples

Just 2 at the moment

I'd like to request that people add some more, in order to capture the kinds of things they intend to use the dataframe Api for

@kkraus14 and @shwina I've seen you both make examples of dataframe agnostic functions in calls - if you could add such examples to the repo this will make conversations going forwards much easier

Thanks!

PR #294, opened 2023-10-23
Add example of a sklearn like pipeline

cbourjau:
This is an example of how a (possibly) lazy data frame may be use in a sklearn-like pipeline.

The example is motivated by the prospect of a fully lazy, ONNX-based data frame implementation. The concept is that calls to `fit` are eager. They compute some state that is later meant to be transferred into the ONNX graph/model. That transfer happens when calling `transform`. The logic within the `transform` methods is "traced" lazily and the resulting lazy object is then exported to ONNX.

Related to #281 and https://github.com/data-apis/dataframe-api/pull/249#issuecomment-1769213663

### DIFF
--- a/spec/API_specification/examples/03_sklearn_like_pipeline.py
+++ b/spec/API_specification/examples/03_sklearn_like_pipeline.py
@@ -33,8 +33,8 @@ class Scaler:
             scalings = scalings.collect()
 
         self.scalings_ = {
-            # Clarify: Should the return type of `get_value` (lazy /
-            # eager scalar) depend on a previous call to `collect`?
+            # Note: `get_value` returns an implemenation-defined,
+            # duck-typed scalar which may be lazy.
             column_name: scalings.get_column_by_name(column_name).get_value(0)
             for column_name in self.column_names
         }
--- a/spec/API_specification/examples/03_sklearn_like_pipeline.py
+++ b/spec/API_specification/examples/03_sklearn_like_pipeline.py
@@ -1,0 +1,127 @@
+"""
+This is an example of how a (possibly) lazy data frame may be used
+in a sklearn-like pipeline.
+
+The example is motivated by the prospect of a fully lazy, ONNX-based
+data frame implementation. The concept is that calls to `fit` are
+eager. They compute some state that is later meant to be transferred
+into the ONNX graph/model. That transfer happens when calling
+`transform`. The logic within the `transform` methods is "traced"
+lazily and the resulting lazy object is then exported to ONNX.
+""" 
+from __future__ import annotations
+
+from typing import Any, TYPE_CHECKING, Self
+
+from dataframe_api.dataframe_object import DataFrame
+
+#: Dummy type alias for a standard compliant array object
+Array = Any
+
+
+class Scaler:
+    """Apply a standardization scaling factor to `column_names`."""    
+    scalings_: dict[str, float]
+
+    def __init__(self, column_names: list[str]):
+        self.column_names = column_names
+
+    def fit(self, df: DataFrame) -> Self:
+        """Compute scaling factors from given data frame.
+
+        Calling this function requires collecting values.
+        """
+        self.scalings_ = {}
+        for column_name in df.column_names:
+            if not column_name in self.column_names:
+                continue
+            self.scalings_[column_name] = df.get_column_by_name(column_name).std()
+
+        return self
+
+    def transform(self, df: DataFrame) -> DataFrame:
+        """Apply the "trained" scaling values.
+
+        This function is guaranteed to not collect values.
+        """
+        df = copy_df(df)
+        for column_name in df.column_names:
+            if not column_name in self.column_names:
+                continue
+            column = df.get_column_by_name(column_name) / self.scalings_[column_name]
+            df.assign(column)
+
+        return df
+
+class FeatureSelector:
+    """Limit columns to those seen in training including their order."""
+
+    def fit(self, df: DataFrame) -> Self:
+        """Record the observed columns and their order.
+
+        This function is guaranteed to not collect values.
+        """
+        self.columns_ = df.column_names
+        return self
+
+    def transform(self, df: DataFrame) -> DataFrame:
+        """Select and sort the columns as observed in training.
+
+        This function is guaranteed to not collect values.
+        """
+        # FIXME: Does this ensure column order?
+        return df.select(self.columns_)
+
+
+class Pipeline:
+    """Linear pipeline of transformers."""
+
+    def __init__(self, steps: list[Any]):
+        self.steps = steps
+    
+    def fit(self, df: DataFrame) -> Self:
+        """Call filt on the steps of the pipeline subsequently.
+
+        Calling this function may trigger a collection.
+        """
+        for step in self.steps:
+            step.fit(df)
+
+        self.steps_ = self.steps
+        return self
+
+    def transform(self, df: DataFrame) -> DataFrame:
+        """Call transform on all steps of this pipeline subsequently.
+
+        This function is guaranteed to not trigger a collection.
+        """
+        for step in self.steps_:
+            df = step.transform(df)
+
+        return df
+
+
+def copy_df(df: DataFrame):
+    """Create a copy of `df`.
+
+    This is done by converting a data frame into standard-arrays and
+    assembling them back into a new data frame.
+    """
+    dfx = df.__dataframe_namespace__()
+
+    dct = dataframe_to_dict_of_arrays(df)
+    return dfx.dataframe_from_dict(
+        # FIXME: This would require some kind of dtype mapping?
+        {column_name: dfx.column_from_1d_array(arr, dtype=arr.dtype) for column_name, arr in dct.items()}
+    )
+
+
+def dataframe_to_dict_of_arrays(df: DataFrame) -> dict[str, Array]:
+    """Convert the given data frame into an dictionary of standard arrays. """
+    dct = {}
+    dfx = df.__dataframe_namespace__()
+    for column_name in df.column_names:
+        column = df.get_column_by_name(column_name)
+        dct[column_name] = column.to_array_object(column.dtype)
+
+    return dct
--- a/spec/API_specification/examples/03_sklearn_like_pipeline.py
+++ b/spec/API_specification/examples/03_sklearn_like_pipeline.py
@@ -15,9 +15,6 @@ from typing import Any, TYPE_CHECKING, Self
 
 from dataframe_api.dataframe_object import DataFrame
 
-#: Dummy type alias for a standard compliant array object
-Array = Any
-
 
 class Scaler:
     """Apply a standardization scaling factor to `column_names`."""    
@@ -31,11 +28,16 @@ class Scaler:
 
         Calling this function requires collecting values.
         """
-        self.scalings_ = {}
-        for column_name in df.column_names:
-            if not column_name in self.column_names:
-                continue
-            self.scalings_[column_name] = df.get_column_by_name(column_name).std()
+        scalings = df.select(self.column_names).std()
+        if hasattr(scalings, 'collect'):
+            scalings = scalings.collect()
+
+        self.scalings_ = {
+            # Clarify: Should the return type of `get_value` (lazy /
+            # eager scalar) depend on a previous call to `collect`?
+            column_name: scalings.get_column_by_name(column_name).get_value(0)
+            for column_name in self.column_names
+        }
 
         return self
 
@@ -44,12 +46,12 @@ class Scaler:
 
         This function is guaranteed to not collect values.
         """
-        df = copy_df(df)
         for column_name in df.column_names:
             if not column_name in self.column_names:
                 continue
             column = df.get_column_by_name(column_name) / self.scalings_[column_name]
-            df.assign(column)
+            # Note: `assign` is not in-place
+            df = df.assign(column)
 
         return df
 
@@ -69,7 +71,7 @@ class FeatureSelector:
 
         This function is guaranteed to not collect values.
         """
-        # FIXME: Does this ensure column order?
+        # Note: This assumes that select ensures the column order.
         return df.select(self.columns_)
 
 
@@ -80,7 +82,7 @@ class Pipeline:
         self.steps = steps
     
     def fit(self, df: DataFrame) -> Self:
-        """Call filt on the steps of the pipeline subsequently.
+        """Call fit on the steps of the pipeline subsequently.
 
         Calling this function may trigger a collection.
         """
@@ -100,28 +102,3 @@ class Pipeline:
 
         return df
 
-
-def copy_df(df: DataFrame):
-    """Create a copy of `df`.
-
-    This is done by converting a data frame into standard-arrays and
-    assembling them back into a new data frame.
-    """
-    dfx = df.__dataframe_namespace__()
-
-    dct = dataframe_to_dict_of_arrays(df)
-    return dfx.dataframe_from_dict(
-        # FIXME: This would require some kind of dtype mapping?
-        {column_name: dfx.column_from_1d_array(arr, dtype=arr.dtype) for column_name, arr in dct.items()}
-    )
-
-
-def dataframe_to_dict_of_arrays(df: DataFrame) -> dict[str, Array]:
-    """Convert the given data frame into an dictionary of standard arrays. """
-    dct = {}
-    dfx = df.__dataframe_namespace__()
-    for column_name in df.column_names:
-        column = df.get_column_by_name(column_name)
-        dct[column_name] = column.to_array_object(column.dtype)
-
-    return dct
--- a/spec/API_specification/examples/07_sklearn_like_pipeline.py
+++ b/spec/API_specification/examples/07_sklearn_like_pipeline.py
@@ -1,4 +1,5 @@
-"""
+"""Scikit-learn-like pipeline.
+
 This is an example of how a (possibly) lazy data frame may be used
 in a sklearn-like pipeline.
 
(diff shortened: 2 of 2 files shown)


spec/API_specification/examples/03_sklearn_like_pipeline.py
```diff
@@ -0,0 +1,127 @@
...
+        """Apply the "trained" scaling values.
+
+        This function is guaranteed to not collect values.
+        """
+        df = copy_df(df)
+        for column_name in df.column_names:
+            if not column_name in self.column_names:
+                continue
+            column = df.get_column_by_name(column_name) / self.scalings_[column_name]
+            df.assign(column)
```
MarcoGorelli, line 52: `assign` isn't in-place, this should be `df = df.assign(column)`

spec/API_specification/examples/03_sklearn_like_pipeline.py
```diff
@@ -0,0 +1,127 @@
...
+        """
+        self.columns_ = df.column_names
+        return self
+
+    def transform(self, df: DataFrame) -> DataFrame:
+        """Select and sort the columns as observed in training.
+
+        This function is guaranteed to not collect values.
+        """
+        # FIXME: Does this ensure column order?
```
MarcoGorelli, line 72: yes I think it should

spec/API_specification/examples/03_sklearn_like_pipeline.py
```diff
@@ -0,0 +1,127 @@
...
+
+        return df
+
+
+def copy_df(df: DataFrame):
+    """Create a copy of `df`.
+
+    This is done by converting a data frame into standard-arrays and
+    assembling them back into a new data frame.
+    """
```
MarcoGorelli, line 109: not all datatypes are supported by the array api. in particular, datetimes aren't

I think you don't need this though, because `assign` isn't inplace

spec/API_specification/examples/03_sklearn_like_pipeline.py
```diff
@@ -0,0 +1,127 @@
...
+    def fit(self, df: DataFrame) -> Self:
+        """Compute scaling factors from given data frame.
+
+        Calling this function requires collecting values.
+        """
+        self.scalings_ = {}
+        for column_name in df.column_names:
+            if not column_name in self.column_names:
+                continue
+            self.scalings_[column_name] = df.get_column_by_name(column_name).std()
```
MarcoGorelli, line 38: I think you could evaluate these all together (and so possibly in parallel);
```python
scalings = df.select(self.column_names).std()
if hasattr(scalings, 'collect'):
    scalings = scalings.collect()

self.scalings_ = {column_name: scalings.get_column_by_name(column_name).get_value(0) for column_name in self.column_names}
```

MarcoGorelli (changes requested):
thanks @cbourjau , much appreciated!

spec/API_specification/examples/03_sklearn_like_pipeline.py
cbourjau: Do you think there should be a standardized way to get to the "physical" (polars parlance) data type of such columns? I believe that this interoperability is rather important, isn't it?

MarcoGorelli:
Before I give my opinion, just to understand better - could you clarify what you mean by lazy scalar please? How would it behave? Could you give an example of how would you use it?

spec/API_specification/examples/03_sklearn_like_pipeline.py
cbourjau: Exactly! Thanks for clarifying!

cbourjau:
I am coming to the data frame API with a point of view that it can be implemented on top of an array-api implementation. In such a latter implementation, there may be lazy tensors. A lazy tensor of rank 0 is what I mean when I say "lazy scalar".

An example of its usage can be seen in [this](https://github.com/data-apis/dataframe-api/blob/main/spec/API_specification/examples/01_standardise_columns.py#L15) example under the assumption that `my_dataframe_agnostic_function` is compatible with lazy implementations of the standard.

MarcoGorelli:
> An example of its usage can be seen in [this](https://github.com/data-apis/dataframe-api/blob/main/spec/API_specification/examples/01_standardise_columns.py#L15) example under the assumption that my_dataframe_agnostic_function is compatible with lazy implementations of the standard.

exactly what I was just thinking about 😄 In that case, I don't think we even need scalars. Your thoughts on https://github.com/data-apis/dataframe-api/issues/297 would be much appreciated if possible

spec/API_specification/examples/03_sklearn_like_pipeline.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def transform(self, df: DataFrame) -> DataFrame:
+        """Apply the "trained" scaling values.
+
+        This function is guaranteed to not collect values.
+        """
+        for column_name in df.column_names:
+            if not column_name in self.column_names:
+                continue
+            column = df.get_column_by_name(column_name) / self.scalings_[column_name]
```
MarcoGorelli, line 52: @kkraus14 just to clarify about scalar issue - should this be written as
```python
column = df.get_column_by_name(column_name) / float(self.scalings_[column_name])
```
instead, so that `__float__` is called and each implementation knows what to do (e.g. trigger compute, raise, return eager value, ...)?

kkraus14: No, that would force materialization unnecessarily. `self.scalings_` items get defined via `scalings.get_column_by_name(column_name).get_value(0)` which could return a ducktyped lazy scalar, and then this division could be lazy as well.

I.E. how this works in cuDF today is that the `get_value(0)` call returns a `cudf.Scalar` object which lives on the GPU, and then the `__div__` operation with the column can use that on-GPU value directly.

Adding `float(...)` requires returning a Python float, which then prevents doing things like the above.

MarcoGorelli: The issue I have is that I would typically do:
- `fit` in one environment
- `predict` in a different environment (e.g. on a smartwatch)

If the "something triggers materialization" happens during `predict`, then the expression is going to fail, because the dataframe which is meant to be used for training isn't available any more

What I want to do is:
- `fit` in one environment, and force materialisation so I can export the model weights
- `predict` in a different environment, using the model weights calculated during the `fit` phase

The way the code is written in this PR achieves this by doing
```python
        if hasattr(scalings, 'collect'):
            scalings = scalings.collect()
```
during the `fit` phase. And we may have to be OK with this if we can't agree on a solution which can be part of the standard

cbourjau: Our use case is close to what you describe, @MarcoGorelli :

We train our models with an eager library (actually pandas at this point). Naturally, this materialized the weights computed during the training. We then want to use our lazy data frame library to "trace" the inference-related code (i.e `Pipeline.transform`).

Is it possible that we are essentially revolving around the following question: "What is the blessed way to get data out of a standard data frame - independently of the question of whether it is a lazy or eager implementation". Right now, I'd say that is the `to_array` interface, but how would we deal with non-array api data? Maybe solving this question will help us here, too?

MarcoGorelli: > Right now, I'd say that is the to_array interface

this doesn't necessarily get the data out, e.g.

```python
In [2]: pl.LazyFrame({'a': [1,2,3]}).to_numpy()
AttributeError                            Traceback (most recent call last)
Cell In[2], line 1
----> 1 pl.LazyFrame({'a': [1,2,3]}).to_numpy()

AttributeError: 'LazyFrame' object has no attribute 'to_numpy'
```

For now I'm having to raise in `dataframe-api-compat` unless the user has called `.collect` - but I'm hoping we can find a standardised solution to that

kkraus14: If you're talking about running different things in two different environments, I assume that means running them in different processes at different times, so that would involve serialization of whatever you want to store. I don't think we've touched on this as of yet and I'm not sure what kind of behaviors we would want.

I.E. with our scalars using `float` and `int` typing (assumed to allow ducktyping), does that mean we need to make `pickle.dumps(...)` compatible with pickling standard float / int objects?

MarcoGorelli: I don't need to save the dataframe, just the fitted model, e.g. https://xgboost.readthedocs.io/en/stable/tutorials/saving_model.html

In order to be able to fit (and then export) an xgboost model, the training data needs to be materialised - that's all I'm trying to do

rgommers: > the training data needs to be materialised - that's all I'm trying to do

I think @kkraus14's point is correct here - in the end _something_ gets serialized (model weights of some sort). So after calling a `result = .fit(...)`, which can in principle still be lazy, you have a `save_on_disk(result)`, and that is the point at which you need the actual values. The `save_on_disk` call may be the thing that actually causes all of the compute to run (either explicitly if needed by the df library, or implicitly) and then the final result to be materialized.

Now `save_on_disk` type functionality is not in scope for the standard. So it's very well possible that the code will look like:
```python
def fit(df, ...):
    # do a ton of stuff that is 100% compatible with both eager and lazy

def store_fitted_model(result):
    # this is library-specific. also note that if this lives in end user code (as is likely),
    # the if-else below isn't necessary, because the end user writes this part only for
    # one specific dataframe library
   if is_dask(result):
        save_on_disk(result.compute())
   elif is_polars(result):
        save_on_disk(result.collect())
   else:
        save_on_disk(result)

data = my_df_lib.load('data_on_disk.zip')
result = fit(data)
store_fitted_model(result)
```

You can further imagine that `fit` is in a library while data loading/storing and calling `fit` is in end user code. Hence, `fit` does _not_ need any execution triggers within its function body here in principle.

MarcoGorelli: Sorry it's not clear to me why you're doing `result.compute` / `result.collect`, as `result` here is the output of `model.fit(x_train, y_train)` and therefore is an instance of a model (e.g. an xgboost model / scikit-learn model), not a dataframe

rgommers: Okay, then `result.weight.collect()`, or whatever is the actual data structure you need values in to transport them to _outside of the dataframe library_.

MarcoGorelli: I think this is the part which we might not be agreeing on:
```python
def fit(df, ...):
    # do a ton of stuff that is 100% compatible with both eager and lazy
```

I'd like to bring attention to the case when `fit` can't be done in a way that's 100% compatible with both eager and lazy - for example, if I need a C library to work with the underlying values. Or, in the concrete case of `skrub`, if `fit` defers to some other library which expects a numpy array as input.
[here](https://github.com/skrub-data/skrub/blob/a8af7d5869d7a8084c46b013efb4d6a1bd88182c/skrub/_similarity_encoder.py#L304-L425)'s a concrete example in which they defer to `CountVectorizer` within `SimilarityEncoder.fit`, which expects a numpy array

In this case, how could `fit` be written in a way that is 100% compatible with both eager and lazy?

MarcoGorelli: > `result.weights.collect()`

also, the weights aren't necessarily storable in a dataframe structure (e.g. the learned parameters from xgboost)

rgommers: >  for example, if I need a C library to work with the underlying values

That's still the same thing - you are at that point calling something _outside of the dataframe library_. At that point you're doing interop - serialization, converting to a foreign data structure (whether via interchange protocol, `.to_array_obj`, or whatever else). It's clear that exactly at the point where you're leaving the dataframe world, and go to something that isn't aware of your lazy constructs, you may need the materialized values. Remember that in the ONNX case there's even a lazy array, so it just depends - there are clear points where you need to have values. But calling `.fit()` or even "calling a C library" are not that (a C library is a bit of a distraction here, it must have a Python API/wrapper somehow, and whether that Python API does or does not support lazy mode is what matters here).


> [here](https://github.com/skrub-data/skrub/blob/a8af7d5869d7a8084c46b013efb4d6a1bd88182c/skrub/_similarity_encoder.py#L304-L425)'s a concrete example in which they defer to `CountVectorizer` within `SimilarityEncoder.fit`, which expects a numpy array

Same answer - _in case_ you need numpy, the exact Python API call where you can't stay lazy is `.to_array`.

> also, the weights aren't necessarily storable in a dataframe structure

Same answer again. If it's for example an XGBoost-native data structure, there's an interop point like `__dataframe__` or `.to_array` you're hitting.

MarcoGorelli: thanks for looking into this, we may be getting closer

>  in case you need numpy, the exact Python API call where you can't stay lazy is .to_array.

right, if a `DataFrame` is backed by an eager Polars DataFrame, then I can call `to_array` and convert to numpy. But if it's backed by a Polars LazyFrame, then I can't, `LazyFrame.to_numpy` would raise - a Polars LazyFrame needs to be collected before it can be converted to numpy.
So how would one force/hint that collection to happen?
(I don't really want it to happen automatically within `to_array`)

rgommers: Yes, I think we're aligned there - that (`to_array`) is where the hinting is needed.

This issue would be much easier to discuss than higher-level hand-wavy arguments about things like `.fit`, because that's not really correct/applicable. The only thing that matters is which exact Python APIs one is calling - and `to_array` is the most obvious one where one leaves the dataframe standard and hence that _may but doesn't have to_ require materialization.

MarcoGorelli: i think the hinting may be needed before to_array

anyway, going out now, happy to speak on monday - I'll prepare the skrub example so we have something concrete to work with, I'm sure we can find a solution - have a nice weekend!

rgommers: > i think the hinting may be needed before to_array

Yes, I understand you want/need to spell it like `.collect().to_array()`.

> have a nice weekend!

Thanks - you too!

MarcoGorelli: I've updated the skrub draft PR, comments on how you're all expecting the dataframe api to be used would be very welcome: https://github.com/skrub-data/skrub/pull/786

cbourjau: Thanks for this IMHO illuminating discussion! I think we are getting closer to finding the right mental model to talk about this problem! Rather than a question of "lazy" vs. "eager" it appears to me to be (partially) a more generic question about how to leave the "data frame standard world". I have a few more observations which I would like to point out:

#### Duck typing

I am worried that the details of what that actually entails might be too fuzzy for a standard. Is there a definitive list of methods that a "Python scalar" must implement?

#### `to_array` and the relationship between the data frame API and the array API standard

The `to_array` function seems like an obvious exit point from the dataframe standard, but it begs the question of how the dataframe and the array API standards are to interact. Is any given implementation of the data frame standard somehow generic over all array API implementations? Or is it rather a predefined relationship where any dataframe implementation specifies exactly one array API implementation? 

As @rgommers points out, `to_array` really only kicks the can of how to hint/force materialization down the road into the array API standard. I think this is a feature rather than a bug, since conceptually the same problem exists on that level of abstraction.

#### How to deal with data types that are unsupported by the array API standard?
If there is consensus that `to_array` should be part of the standard in any case then maybe it could be possible to only allow leaving the data frame standard via array API standard data type for now without limiting future design space. Or do you see better alternatives here?

MarcoGorelli: totally agree on the second part - I've opened #305 to discuss

> Or is it rather a predefined relationship where any dataframe implementation specifies exactly one array API implementation?

I think it's this - each implementation can choose which array library to defer to when someone calls `to_array`. For cudf it would probably be cupy, for pandas/polars numpy

Regarding datatypes - yeah `to_array` only supports datatypes from the array api standard

kkraus14: > > Or is it rather a predefined relationship where any dataframe implementation specifies exactly one array API implementation?
> 
> I think it's this - each implementation can choose which array library to defer to when someone calls `to_array`. For cudf it would probably be cupy, for pandas/polars numpy
> 
> Regarding datatypes - yeah `to_array` only supports datatypes from the array api standard

I think where things get a bit fuzzy is for a library like Dask which has Dask Array and can in theory keep things lazy between DataFrames and Arrays, so `to_array` shouldn't necessarily force materialization, but for libraries like Polars Lazy it seems like we don't want `to_array` to implicitly make materialization happen either. We're left in a situation where the only "correct" behavior would be to allow throwing in `to_array` but that's quite a poor developer experience.

rgommers: I don't think that's the only option. I think we want a `.may_execute()` type method (ignore the exact method name here), which in the example above (and also @cbourjau's ONNX library which has lazy arrays) can simply be do-nothing, while Polars could alias that to `.collect()`. It would address the concern about implicit materialization, making it explicit.

It wouldn't be super ergonomic to write it as `.may_execute().to_array()`, but probably better than throwing - which isn't a solution really.

This solution is kinda circling back to a small part of Marco's lazy/eager design, without the separate classes or expression objects.

Other thought on Polars there: it could also choose to do an optimization pass, where e.g. it could figure out for a block of code with multiple `may_execute`'s which ones have to collect and which ones don't. That'd actually be generically useful outside of dataframe API support, because the manual "must execute" imperative nature of `.collect` is not so easy to write optimally (as we already discovered in another discussion on this repo).

MarcoGorelli: `may_execute` sounds fine, thanks

btw yesterday I saw a colleague write this

> Has anyone seen hvplot causing 3 identical computes when working off of a dask dataframe? (version 0.9.0) I dont know if I should be calling something differently or if this is a bug?

. If I was -1 on automated collection before, now I'm more like -10 😄

MarcoGorelli: Note that for `skrub.<>.fit`, even Dask would throw an error - I've added an example of this in https://github.com/data-apis/dataframe-api/pull/307, as well as a proposal

cbourjau: >> Or is it rather a predefined relationship where any dataframe implementation specifies exactly one array API implementation?

> I think it's this - each implementation can choose which array library to defer to when someone calls to_array. For cudf it would probably be cupy, for pandas/polars numpy

If we have a 1:1 relationship between the dataframe and the array API standard then we have successfully kicked the can down the road (i.e. tech-stack) into the array API standard. Calling `to_array` will always trigger materialization if the associated array-API implementation is eager. If the associated array-API implementation is lazy, then it will remain lazy. Would this be acceptable for the Pandas/Polars-backed implementations? I'm afraid I don't quite see the intention behind `may_execute`?

The above-described semantics would work well with ONNX as far as the dataframe API is concerned. We are left with the same question of explicit materialization on the array API level, though.

MarcoGorelli: > Calling to_array will always trigger materialization if the associated array-API implementation is eager. [...] Would this be acceptable for the Pandas/Polars-backed implementations?

I'm afraid not, sorry: https://github.com/data-apis/dataframe-api/pull/307#issuecomment-1794535943

MarcoGorelli (commented):
I think we can update based on recent scalars / `persist` discussions. But first, just wanted to check something

Realistically, would you want the output of `.fit` to be lazy?

I think you'd typically want it to be guaranteed to not be lazy. In my experience in data science, what was typical was:
- fit model once (this can be expensive)
- predict / transform multiple times, on multiple datasets (this is typically cheaper)

So, to be honest, I would want `fit` to evaluate everything it needs to, so that `predict` / `transform` can be cheap

Does this match your experience?

If so, maybe we should write
```python
       self.scalings_ = {
            column_name: float(scalings.get_column_by_name(column_name).get_value(0))
            for column_name in self.column_names
        }
```
, in which case we'd need to add `__float__` to the Scalar class. Assuming you even want a float

Adding `persist` inside `fit` isn't really ideal, because it means that `scalings.col(column_name).get_value(0)` will be repeatedly evaluated inside each `transform` call

kkraus14:
> Adding `persist` inside `fit` isn't really ideal, because it means that `scalings.col(column_name).get_value(0)` will be repeatedly evaluated inside each `transform` call

Could you do `scalings.col(column_name).get_value(0).persist()`? I imagine we'd need the persist method on every class type we have.

MarcoGorelli:
Why persist? Persist is a hint which tells the library "don't repeat computation prior to this point"

I'm trying to get the Python value out, I think we need something like `.collect`

MarcoGorelli:
I've pushed some commits to update this based on other recent API changes, anyone have objections?

MarcoGorelli:
approved during the call, thanks @cbourjau !

MarcoGorelli (approved):

PR #298, opened 2023-10-24
Note that comparisons between columns from different dataframes is outside the standard

MarcoGorelli:
This has come up a few times

https://github.com/data-apis/dataframe-api/pull/249#issuecomment-1769254879

> Are folks open to simply saying that the operation is undefined by the existing standard? Implementations can choose to raise

https://github.com/data-apis/dataframe-api/issues/287#issuecomment-1777097662

> On way to do [Column.mean() whilst not having to materialise more times than necessary when filtering] is for the column to keep track of its originating DataFrame:

---

so let's document that it's outside the standard?

### DIFF
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -211,6 +211,12 @@ class Column:
         Returns
         -------
         Column
+
+        Notes
+        -----
+        If `other` is a `Column`, then the Standard only supports the case when `other`
+        and `self` originated from the same dataframe. If they didn't, then behaviour
+        is unspecified and may vary across implementations.
         """
 
     def __ne__(self: Column, other: Column | Scalar) -> Column:  # type: ignore[override]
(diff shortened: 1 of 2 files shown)


rgommers:
Thanks Marco. This PR is useful in identifying all (?) the potentially problematic methods and operator support. It makes it more concrete where we have tension between current eager and lazy libraries. 

> Are folks open to simply saying that the operation is undefined by the existing standard? Implementations can choose to raise

I'm not sure this is actually a good way to go. I'd like to explore a little more if we really can't do better than this, because it's at best unsatisfactory - as commented on in https://github.com/data-apis/dataframe-api/pull/249/#issuecomment-1781355343.

MarcoGorelli:
It's not just eager-vs-lazy, but also that eager libraries arguably shouldn't be doing more passes through their data then they need to https://github.com/data-apis/dataframe-api/issues/287#issuecomment-1777097662

MarcoGorelli:
From today's call: `Column`s may have an "owner" `DataFrame`, need to define expectations around how they behave (e.g. same owner, different owners, no owner)

MarcoGorelli:
> From today's call: `Column`s may have an "owner" `DataFrame`, need to define expectations around how they behave (e.g. same owner, different owners, no owner)

taken forwards in #310

MarcoGorelli:
reopening as maybe this is an alternative, see https://github.com/data-apis/dataframe-api/pull/310#discussion_r1379077940

PR #301, opened 2023-10-26
Fixup typing of Namespace

MarcoGorelli:
I'd missed a couple of details in my implementation, the more accurate and usable the typing is the more that can be avoided


MarcoGorelli:
merging as this only affects typing

PR #302, opened 2023-10-26
remove `names` from dataframe_from_2d_array

MarcoGorelli:
The keys of `dtypes` already give the column names, and Python dicts are ordered

If anyone needs to reorder the names after construction, they can put `.select` after this method anyway

But having to always provide both `dtypes` and `names` is quite annoying from a user perspective

### DIFF
--- a/spec/API_specification/dataframe_api/__init__.py
+++ b/spec/API_specification/dataframe_api/__init__.py
@@ -137,7 +137,7 @@ def column_from_1d_array(array: Any, *, dtype: DType, name: str = '') -> Column:
     """
     ...
 
-def dataframe_from_2d_array(array: Any, *, dtypes: Dict[str, Any]) -> DataFrame:
+def dataframe_from_2d_array(array: Any, *, schema: Dict[str, DType]) -> DataFrame:
     """
     Construct DataFrame from 2D array.
 
--- a/spec/API_specification/dataframe_api/__init__.py
+++ b/spec/API_specification/dataframe_api/__init__.py
@@ -135,7 +135,7 @@ def column_from_1d_array(array: Any, *, dtype: DType, name: str = '') -> Column:
     """
     ...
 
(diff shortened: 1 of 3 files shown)


spec/API_specification/dataframe_api/__init__.py
```diff
@@ -135,7 +135,7 @@ def column_from_1d_array(array: Any, *, dtype: DType, name: str = '') -> Column:
     """
     ...
 
-def dataframe_from_2d_array(array: Any, *, names: Sequence[str], dtypes: Mapping[str, Any]) -> DataFrame:
+def dataframe_from_2d_array(array: Any, *, dtypes: Mapping[str, Any]) -> DataFrame:
```
kkraus14, line 138: Does `Mapping` guarantee ordering? As far as I can tell it only guarantees that `__getitem__`, `__len__`, and `__iter__` are supported, but doesn't guarantee any specific order for `__iter__`?

spec/API_specification/dataframe_api/__init__.py
```diff
@@ -137,7 +137,7 @@ def column_from_1d_array(array: Any, *, dtype: DType, name: str = '') -> Column:
     """
     ...
 
-def dataframe_from_2d_array(array: Any, *, names: Sequence[str], dtypes: Mapping[str, Any]) -> DataFrame:
+def dataframe_from_2d_array(array: Any, *, dtypes: Dict[str, Any]) -> DataFrame:
```
kkraus14, line 140: Nitpick: but we have a `schema` function that is logically equivalent to what we expect here. Maybe rename this parameter to schema to be consistent?

Seeing `dtypes` makes me think I can pass a sequence.

kkraus14 (approved):

spec/API_specification/dataframe_api/__init__.py
MarcoGorelli: you're right thanks for spotting this

PR #307, opened 2023-10-31
Add DataFrame.persist, and notes on execution model

MarcoGorelli:
For now I'm keeping `Column.to_array` out of it - once we sort out #298 , we can add that too

### DIFF
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -980,15 +980,36 @@ class DataFrame(Protocol):
     
     def persist(self) -> Self:
         """
-        Hint that execution may be triggered, depending on the implementation.
+        Hint to that computation prior to this point should not be repeated.
 
         This is intended as a hint, rather than as a directive. Implementations
         which do not separate lazy vs eager execution may ignore this method and
-        treat it as a no-op. Likewise for implementations which support automated
-        execution.
+        treat it as a no-op.
 
         .. note::
-            This method may force execution. If necessary, it should be called
+            This method may trigger execution. If necessary, it should be called
             at most once per dataframe, and as late as possible in the pipeline.
+
+            For example, do this
+
+            .. code-block:: python
+
+                df: DataFrame
+                df = df.persist()
+                features = []
+                for column_name in df.column_names:
+                    if df.col(column_name).std() > 0:
+                        features.append(column_name)
+
+            instead of this:
+
+            .. code-block:: python
+
+                df: DataFrame
+                features = []
+                for column_name in df.column_names:
+                    # Do NOT do this!
+                    if df.persist().col(column_name).std() > 0:
+                        features.append(column_name)
         """
         ...
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -67,7 +67,7 @@ class DataFrame(Protocol):
 
         Notes
         -----
-        To be guaranteed to run across all implementations, :meth:`may_execute` should
+        To be guaranteed to run across all implementations, :meth:`persist` should
         be executed at some point before calling this method.
         """
         ...
@@ -978,7 +978,7 @@ class DataFrame(Protocol):
         """
         ...
     
-    def may_execute(self) -> Self:
+    def persist(self) -> Self:
         """
         Hint that execution may be triggered, depending on the implementation.
 
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -64,6 +64,11 @@ class DataFrame(Protocol):
     def shape(self) -> tuple[int, int]:
         """
         Return number of rows and number of columns.
+
+        Notes
+        -----
+        To be guaranteed to run across all implementations, :meth:`may_execute` should
+        be executed at some point before calling this method.
         """
         ...
 
@@ -928,6 +933,9 @@ class DataFrame(Protocol):
         may choose to return a numpy array (for numpy prior to 2.0), with the
         understanding that consuming libraries would then use the
         ``array-api-compat`` package to convert it to a Standard-compliant array.
+
+        To be guaranteed to run across all implementations, :meth:`may_execute` should
+        be executed at some point before calling this method.
         """
     
     def join(
@@ -970,5 +978,20 @@ class DataFrame(Protocol):
         ValueError
             If, apart from `left_on` and `right_on`, there are any column names
             present in both `self` and `other`.
+        """
+        ...
+    
+    def may_execute(self) -> Self:
+        """
+        Hint that execution may be triggered, depending on the implementation.
+
+        This is intended as a hint, rather than as a directive. Implementations
+        which do not separate lazy vs eager execution may ignore this method and
+        treat it as a no-op. Likewise for implementations which support automated
+        execution.
+
+        .. note::
+            This method may force execution. If necessary, it should be called
+            at most once per dataframe, and as late as possible in the pipeline.
         """
         ...
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -931,7 +931,7 @@ class DataFrame(Protocol):
         ...
 
     def persist(self) -> Self:
-        """Hint to that computation prior to this point should not be repeated.
+        """Hint that computation prior to this point should not be repeated.
 
         This is intended as a hint, rather than as a directive. Implementations
         which do not separate lazy vs eager execution may ignore this method and
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -67,7 +67,7 @@ class DataFrame(Protocol):
 
         Notes
         -----
-        To be guaranteed to run across all implementations, :meth:`maybe_execute` should
+        To be guaranteed to run across all implementations, :meth:`may_execute` should
         be executed at some point before calling this method.
         """
         ...
@@ -933,9 +933,6 @@ class DataFrame(Protocol):
         may choose to return a numpy array (for numpy prior to 2.0), with the
         understanding that consuming libraries would then use the
         ``array-api-compat`` package to convert it to a Standard-compliant array.
-
-        To be guaranteed to run across all implementations, :meth:`maybe_execute` should
-        be executed at some point before calling this method.
         """
     
     def join(
@@ -981,7 +978,7 @@ class DataFrame(Protocol):
         """
         ...
     
-    def maybe_execute(self) -> Self:
+    def may_execute(self) -> Self:
         """
         Hint that execution may be triggered, depending on the implementation.
 
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -67,7 +67,7 @@ class DataFrame(Protocol):
 
         Notes
         -----
-        To be guaranteed to run across all implementations, :meth:`may_execute` should
+        To be guaranteed to run across all implementations, :meth:`maybe_execute` should
         be executed at some point before calling this method.
         """
         ...
@@ -934,7 +934,7 @@ class DataFrame(Protocol):
         understanding that consuming libraries would then use the
         ``array-api-compat`` package to convert it to a Standard-compliant array.
 
-        To be guaranteed to run across all implementations, :meth:`may_execute` should
+        To be guaranteed to run across all implementations, :meth:`maybe_execute` should
         be executed at some point before calling this method.
         """
     
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -62,9 +62,7 @@ class DataFrame(Protocol):
         ...
 
     def shape(self) -> tuple[int, int]:
-        """
-        Return number of rows and number of columns.
-        """
+        """Return number of rows and number of columns."""
         ...
 
     def group_by(self, *keys: str) -> GroupBy:
@@ -931,10 +929,9 @@ class DataFrame(Protocol):
             present in both `self` and `other`.
         """
         ...
-    
+
     def persist(self) -> Self:
-        """
-        Hint to that computation prior to this point should not be repeated.
+        """Hint to that computation prior to this point should not be repeated.
 
         This is intended as a hint, rather than as a directive. Implementations
         which do not separate lazy vs eager execution may ignore this method and
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -981,7 +981,7 @@ class DataFrame(Protocol):
         """
         ...
     
-    def may_execute(self) -> Self:
+    def maybe_execute(self) -> Self:
         """
         Hint that execution may be triggered, depending on the implementation.
 
--- a/spec/design_topics/execution_model.md
+++ b/spec/design_topics/execution_model.md
@@ -45,41 +45,5 @@ for such an operation to be executed:
   TypeError: Trying to convert dd.Scalar<gt-bbc3..., dtype=bool> to a boolean value. Because Dask objects are lazily evaluated, they cannot be converted to a boolean value or used in boolean conditions like if statements. Try calling .compute() to force computation prior to converting to a boolean value or using in a conditional statement.
   ```
 
-## Solution: DataFrame.persist
-
-The Dataframe API has a `DataFrame.persist` for addressing the above. We can use it to rewrite the code above
-as follows:
-```python
-df: DataFrame
-df = df.persist()
-features = []
-for column_name in df.column_names:
-    if df.col(column_name).std() > 0:
-        features.append(column_name)
-return features
-```
-
-Note that `persist` is to be interpreted as a hint, rather than as a directive -
-the implementation itself may decide
-whether to force execution at this step, cache results, defer it to later and cache
-results when they reach the current node, or ignore it.
-
-Operations which require `DataFrame.persist` to have been called at some prior
-point are:
-- `DataFrame.shape`
-- calling `bool`, `int`, or `float` on a scalar 
-
-Note how in the above example, `DataFrame.persist` was called only once,
-and as late as possible.
-Conversely, the "wrong" way to execute the above would be:
-
-```python
-df: DataFrame
-features = []
-for column_name in df.column_names:
-    # Do NOT do this!
-    if df.persist().col(column_name).std() > 0:
-        features.append(column_name)
-return features
-```
-as that will potentially re-trigger the same execution multiple times.
+Whether such computation succeeds or raises is currently not defined by the Standard and may vary across
+implementations.
--- a/spec/design_topics/execution_model.md
+++ b/spec/design_topics/execution_model.md
@@ -38,40 +38,13 @@ for such an operation to be executed:
      ...:     print('scalar is positive')
      ...:
   ---------------------------------------------------------------------------
-  TypeError                                 Traceback (most recent call last)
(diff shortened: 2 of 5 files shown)


cbourjau:
Would you mind elaborating a bit more on what the benefit of `may_evaluate` is compared to explicitly allowing certain functions such as `shape` (unless we reuse the semantics from the array-API standard's `shape` [function](https://data-apis.org/array-api/draft/API_specification/generated/array_api.array.shape.html)?) and `to_array` to optionally materialize?

MarcoGorelli:
Sure - anything related to automatic execution is going to result in people accidentally double-computing things

Quick example:
```python
df: DataFrame
features = df.drop_columns('target').to_array()  # triggers the whole DAG behind `df`
target = df.col('target').to_array()  # also triggers the whole DAG behind `df`
```
as opposed to
```python
df: DataFrame
df = df.maybe_execute()
features = df.drop_columns('target').to_array()
target = df.col('target').to_array()
```

In the first example, Polars would push down the `'target'` selection

cbourjau:
I'm afraid I still don't quite see it, sorry! How would you implement `to_array` in your second example to take advantage of the `maybe_execute` hint assuming that (a) `to_array` returns an array-API compliant array and (b) that array is a numpy array for the pandas-backed-dataframe implementation in question?

Would there be some cache associated with `df` after the `maybe_execute` call? If so, would that not deteriorate the user experience with respect to memory consumption? I.e. the user may expect the memory of the `features` array to be GC'ed after `features` went out of scope, but instead it would live for as long as `df` does.

MarcoGorelli:
Thanks for asking, this will help clarify things

So, let's consider the following cases

## case 1: lazy dataframe, with lazy array counterpart, which requires computation for `bool(col.std())`

You could have `may_execute` just set a boolean flag, which is then ignored for `Column.to_array` and `DataFrame.to_array` but is respected for `Column.std().__bool__()`

Something like:
```
class Scalar:
    def __bool__(self):
        if self.parent_df.may_execute:
            self.parent_df = self.parent_df.compute()
            return bool(value)
        else:
            raise RuntimeError("calling `bool` in a scalar requires the parent df to have been evaluated")

class Column:
    def to_array(self):
        # ignore `self.may_execute` here, as `to_array` doesn't need it
        return self.column.to_array()

    def std(self):
        return Scalar(self.column.std(), parent_df = self.parent_df)
```


## case 2: lazy dataframe, with eager array counterpart, which requires computation for `bool(col.std())`

Could just alias `may_execute` to the underlying `collect` / `compute` call, e.g.
```python
class DataFrame:
    # [...]
    def may_execute(self):
        return DataFrame(self.dataframe.collect())
```

## case 3: eager dataframe, everything's eager

`may_execute` is a no-op
```python
class DataFrame:
    # [...]
    def maybe_execute(self):
        return DataFrame(self.dataframe)
```

spec/design_topics/execution_model.md
```diff
@@ -0,0 +1,83 @@
...
+        features.append(column_name)
+return features
+```
+
+Note that `maybe_evaluate` is to be interpreted as a hint, rather than as a directive -
+the implementation itself may decide
+whether to force execution at this step, or whether to defer it to later.
+For example, a dataframe which can convert to a lazy array could decide to ignore
+`maybe_evaluate` when evaluting `DataFrame.to_array` but to respect it when evaluating
+`float(Column.std())`.
```
cbourjau, line 63: So, `maybe_evaluate` may do:
- Nothing at all
- Nothing at this point but allows later collections when the backend thinks it is expedient
- An immediate collection

What happens to subsequent calls to `df` assuming that a collection did take place? Are they eager or lazy?
```python
df: DataFrame
column_name: str
df = df.maybe_execute()
col = df.col(column_name)
filtered_col = col.filter(col > 42)  # is this computation eager now?
filtered_col.std()
```

spec/design_topics/execution_model.md
```diff
@@ -0,0 +1,83 @@
...
+```python
+df: DataFrame
+features = []
+for column_name in df.column_names:
+    # Do NOT do this!
+    if df.maybe_execute().col(column_name).std() > 0:
+        features.append(column_name)
+return features
+```
+as that will potentially re-trigger the same execution multiple times.
```
cbourjau, line 83: But there are no guarantees here, are there? Given that `maybe_execute` still allows for deferred execution the backend may still re-trigger the same execution multiple times.

cbourjau (commented):
Thanks for the further explanations even though I must say that I am still a bit confused about the semantics of `maybe_execute`. This [example](https://github.com/data-apis/dataframe-api/pull/307#issuecomment-1794535943) benefits from a collection taking place immediately when `maybe_execute` is called. However, the latter example appears to revolve around the idea of allowing a deferred collection when `__bool__` is called. 

A fully lazy implementation would need to error out when calling `__bool__` regardless of the fact if a user called `maybe_execute` or not. A lazy/eager hybrid such as a Polars-backed implementation certainly benefits from explicit "collection hints" if they are computed explicitly where evoked. 

This makes me wonder about the benefit of also using `maybe_execute` to allow subsequent collections compared to simply always allowing those by default.

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -64,6 +64,11 @@ def dataframe(self) -> SupportsDataFrameAPI:
     def shape(self) -> tuple[int, int]:
         """
         Return number of rows and number of columns.
+
+        Notes
+        -----
+        To be guaranteed to run across all implementations, :meth:`maybe_execute` should
+        be executed at some point before calling this method.
```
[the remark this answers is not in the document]
MarcoGorelli: that's right, though `Column` would still be backed by an expression (which is lazy), but the parent dataframe would be eager. you can try this out with https://github.com/data-apis/dataframe-api-compat

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -972,3 +980,18 @@ def join(
...
+        Hint that execution may be triggered, depending on the implementation.
+
+        This is intended as a hint, rather than as a directive. Implementations
+        which do not separate lazy vs eager execution may ignore this method and
+        treat it as a no-op. Likewise for implementations which support automated
+        execution.
+
+        .. note::
+            This method may force execution. If necessary, it should be called
+            at most once per dataframe, and as late as possible in the pipeline.
```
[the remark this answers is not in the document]
MarcoGorelli: if you're using it multiple times, then you're potentially re-executing things

spec/design_topics/execution_model.md
MarcoGorelli: > # is this computation eager now?

It's implementation-dependent. It can stay lazy

What really matters is when you do
```python
bool(filtered_col.std())
```
(which you might trigger via `if filtered_col.std() > 0` - at that point:
- if `maybe_execute` wasn't called previously, this is unsupported by the standard and may vary across implementations
- if `maybe_execute` was called, then libraries supporting eager evaluation should return a result

spec/design_topics/execution_model.md
MarcoGorelli: yes you're right, that's an issue with the suggestion I put in https://github.com/data-apis/dataframe-api/pull/307#issuecomment-1795062767

not sure what to suggest, will think about this

MarcoGorelli: Looks like there's two cases that really need addressing:
- `bool(scalar)` requires computation in all cases
- `to_array` only requires computation in some cases

spec/API_specification/dataframe_api/dataframe_object.py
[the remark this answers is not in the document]
cbourjau: Sure, but there are reasonable cases where that would be what you want, are there not? For example, you may want to collect a dataframe, filter it further in a lazy manner, and then collect it again.

spec/design_topics/execution_model.md
cbourjau: When would you want the first option rather than an implicit default for the latter behavior? It seems rather obvious that `bool(filtered_col.std())` needs to materialize something so it is hardly a surprise to the user at this point. Sure, the user may want to strategically place a `maybe_execute` earlier for performance reasons, but why introduce undefined behavior if they don't?

spec/API_specification/dataframe_api/dataframe_object.py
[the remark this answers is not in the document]
MarcoGorelli: sure but why would you collect it before filtering?

spec/design_topics/execution_model.md
MarcoGorelli: you don't need to introduce undefined behaviour, I just mean that it's undefined by the Dataframe API - the Standard makes no guarantee of what will happen there

kkraus14:
I think `may_execute` is too ambiguous here. It just implies that the dataframe may hit a blocking operation at some point, but doesn't clearly indicate that it should explicitly execute and block or otherwise do something to influence downstream usage. Also, does doing another operation after `may_execute` invalidate the `may_execute` flag?

If we look at lazy execution frameworks (ignoring Ibis since it defers to different backends for actual implementation / execution), they all have a method like `cache` (Polars, Spark) / `persist` (Dask, Spark) that don't block but explicitly instruct that all future usage after that operation will not retrigger computation before that operation.

What would your thoughts be if we added a method like `cache` that is a no-op for eager DataFrames, which gives a better path to avoid performance pitfalls in lazy dataframes, but then generally allowed operations like `to_array` and `bool(Scalar)` to always allow running?

MarcoGorelli:
Initial thought - sounds good! Will try it out in my implementation / skrub and see where this takes us

MarcoGorelli:
> Initial thought - sounds good! Will try it out in my implementation / skrub and see where this takes us

Just tried, and no, `polars.LazyFrame.cache` doesn't persist the dataframe into memory, see https://github.com/pola-rs/polars/issues/2842

Having said that, maybe `persist` could just defer to `.collect` in Polars. If it's clearly documented that it should be called as little and as late as possible, then I'd be OK with that

MarcoGorelli:
Having said that, as far as I can tell,
```python
df: dask.DataFrame
df = df.persist()
```
is equivalent to
```python
df: polars.LazyFrame
df = df.collect().lazy()
```

Is this correct? 🤔 maybe `.persist` is OK then, and we document the list of methods which implicitly force computation

So then
```python
# DataFrame Standard
df: DataFrame
df = df.persist()
features = []
    if df.col(column_name).std() > 0:
        features.append(column_name)
return features
```
would be roughly equivalent to
```python
# polars
df: polars.LazyFrame
df = df.collect().lazy()
features = []
    if df.collect()[column_name].std() > 0:
        features.append(column_name)
return features
```
and
```python
# Dask
df: dask.DataFrame
df = df.persist()
features = []
    if df[column_name].std().compute() > 0:
        features.append(column_name)
return features
```

And maybe that's fine?

kkraus14:
Yea, this is more or less what I had in mind. There's still a footgun if someone doesn't use the method, but it at least gives a standard compliant way for folks to write code that nicely works across both eager and lazy implementations without introducing any implementation burden onto any of the libraries.

MarcoGorelli:
> There's still a footgun if someone doesn't use the method

Sure but we could raise if `persist` hasn't been called earlier?

e.g.
```python
df: DataFrame
features = []
    if df.col(column_name).std() > 0:  # raises, tell you to call `persist` on parent dataframe
        features.append(column_name)
return features
```

Correct way:
```python
df: DataFrame
df = df.persist()
features = []
    if df.col(column_name).std() > 0:
        features.append(column_name)
return features
```

I think the real footgun would be calling `.persist` within a loop:
```python
df: DataFrame
features = []
    if df.persist().col(column_name).std() > 0:
        features.append(column_name)
return features
```
so this is why the "use `.persist` as late and as little as possible" rule should still apply

kkraus14:
> Sure but we could raise if `persist` hasn't been called earlier?
> 
> e.g.
> 
> ```python
> df: DataFrame
> features = []
> for column_name in df.column_names:
>     if df.col(column_name).std() > 0:  # raises, tell you to call `persist` on parent dataframe
>         features.append(column_name)
> return features
> ```

What are the rules for propagating a `persist` status vs invalidating it? I think it would be difficult to have a set of cohesive rules for this.

MarcoGorelli:
It doesn't need to be part of the standard, but an implementation could raise if you try to bring a scalar into Python without having called `persist` at some point between `__dataframe_consortium_standard__` and bringing the scalar into Python

rgommers:
> If we look at lazy execution frameworks (ignoring Ibis since it defers to different backends for actual implementation / execution), they all have a method like `cache` (Polars, Spark) / `persist` (Dask, Spark) that don't block but explicitly instruct that all future usage after that operation will not retrigger computation before that operation.

Trying to catch up here. This seems like a reasonable thing to add, given that all lazy libraries seem to have it. I'm not sure if solves the same problem as `may_execute()`; there are two more or less orthogonal things here:

1. Avoid recomputing expensive calls more than once in lazy implementations
2. Allow writing code where method calls that cannot be kept lazy by a library do work

It looks to me like (1) is solved by `.persist`, while for (2) it's not yet 100% clear to me from the discussion above. It will solve the problem if it returns something that, for Polars, is neither a `LazyFrame` nor an `EagerFrame`. But rather a lazyframe with a few methods that are able trigger execution. I.e. something like:
```python
class LazyFrame:
    ...
    def persist(self) -> PermissiveLazyFrame:
        return self._to_permissive()

class PermissiveLazyFrame(LazyFrame):
    def __bool__(self) -> bool:
        return self.collect().__bool__()

    def __int__(self) -> int:
        return self.collect().__bool__()

    def __float__(self) -> float:
        return self.collect().__bool__()

    def to_array(self) -> numpy.ndarray:
        return self.collect().numpy()

    # add some magic here to ensure that for all other method calls,
    # the return type is PermissiveLazyFrame, not LazyFrame
```
It could be a short implementation, the above is pretty much all that's needed. This would also answer be in line with Marco's answer on when things raise (if `.persist` has been called at least once) - because `LazyFrame` continues raising. Is my understanding correct there?

For the standard, the description of `__bool__` & co would then still be slightly awkward, something like: "returns a boolean scalar; may raise for lazy implementations; is guaranteed not to raise if `.persist()` has been called on the dataframe before unless the library is not able to execute anything eagerly."

MarcoGorelli:
little demo:
```console
$ cat t.py
from __future__ import annotations

import pandas as pd
import polars as pl

from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from dataframe_api.typing import SupportsDataFrameAPI
    from dataframe_api import DataFrame

dfpd = pd.DataFrame({'a': [1, 1, 1], 'b': [4,5,6]})
dfpl = pl.DataFrame({'a': [1, 1, 1], 'b': [4,5,6]})

def this_raises(df_raw: SupportsDataFrameAPI):
    df = df_raw.__dataframe_consortium_standard__(api_version='2023.11-beta')
    features = []
        if df.col(column_name).std() > 0:
            features.append(column_name)
    return features

def this_runs(df_raw: SupportsDataFrameAPI):
    df = df_raw.__dataframe_consortium_standard__(api_version='2023.11-beta')
    df = df.persist()  # type: ignore
    features = []
        if df.col(column_name).std() > 0:
            features.append(column_name)
    return features

def this_runs_but_dont_do_it(df_raw: SupportsDataFrameAPI):
    df = df_raw.__dataframe_consortium_standard__(api_version='2023.11-beta')
    features = []
        if df.persist().col(column_name).std() > 0:  # type: ignore
            features.append(column_name)
    return features
```

Then:
```python
In [1]: this_raises(dfpd)
[...]
In [2]: this_raises(dfpl)
[...]
ValueError: Cannot materialise a lazy dataframe, please call `persist` first

In [3]: this_runs(dfpd)
Out[3]: ['b']

In [4]: this_runs(dfpl)
Out[4]: ['b']

In [5]: this_runs_but_dont_do_it(dfpd)
Out[5]: ['b']

In [6]: this_runs_but_dont_do_it(dfpl)
Out[6]: ['b']

```

spec/API_specification/dataframe_api/dataframe_object.py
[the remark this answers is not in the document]
cbourjau: It is a bit of a constructed example, but maybe you want to do computations on the entire dataframe and also on some subset of it. It would make sense to collect just prior to the first computation on the entire frame so that whatever came before it doesn't have to be recomputed when doing the computation on the subset.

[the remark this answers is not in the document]
MarcoGorelli: something like
```python
df: DataFrame
df = df.persist()
sub_df_1 = df.filter(df.col('a') > 0)
sub_df_2 = df.filter(df.col('a') <= 0)
features_1= []
for column_name in sub_df_1.column_names:
    if sub_df_1.col(column_name).std() > 0:
        features_1.append(column_name)
features_2= []
for column_name in sub_df_2.column_names:
    if sub_df_2.col(column_name).std() > 0:
        features_2.append(column_name)
```
?

You'd still just be calling it once per dataframe - could you show an example of where you'd want to call it twice for the same dataframe?

spec/design_topics/execution_model.md
```diff
@@ -0,0 +1,98 @@
...
+The Dataframe API has a `DataFrame.persist` for addressing the above. We can use it to rewrite the code above
+as follows:
+```python
+df: DataFrame
+df = df.persist()
+features = []
+for column_name in df.column_names:
+    if df.col(column_name).std() > 0:
+        features.append(column_name)
+return features
```
dchigarev, line 59: Isn't calling `.std()` may also be treated as a lazy operation and so prohibit implicit conversion from a `Scalar` to a python boolean? Given the current description of `.persist()` method, I would expect to use it something like this:
```python
df: DataFrame
df = df.persist() # execute the previous computation tree to avoid re-computation in the for-loop
features = []
    if df.col(column_name).std().persist() > 0: # calling '.persist()' again to avoid materialization of the .std() result
        features.append(column_name)
```

MarcoGorelli:
I've updated, and removed the "propagation" part

I think people agreed on everything in this PR

We OK to get this in?

shwina (approved):
Thanks Marco. Apologies for missing yesterday's meeting where perhaps this was discussed.

For what it's worth, I think asking the user of an API to "think lazy" when they don't want/need lazy semantics might make the API difficult to a general audience. But, this API is not for the general audience and I understand that this is the best we can do to support more DataFrame libraries. 

Approving, and thanks for the work here!

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -929,3 +929,38 @@ def join(
             present in both `self` and `other`.
         """
         ...
+
+    def persist(self) -> Self:
+        """Hint to that computation prior to this point should not be repeated.
```
rgommers, line 934:
```suggestion
        """Hint that computation prior to this point should not be repeated.
```

spec/purpose_and_scope.md
```diff
@@ -125,9 +125,10 @@ See the [use cases](use_cases.md) section for details on the exact use cases con
 Implementation details of the dataframes and execution of operations. This includes:
 
 - How data is represented and stored (whether the data is in memory, disk, distributed)
-- Expectations on when the execution is happening (in an eager or lazy way)
+- Expectations on when the execution is happening (in an eager or lazy way), other than `DataFrame.persist`
```
rgommers, line 128: minor: not entirely accurate, since it's only a hint so there is still no "when" prescribed.

How about saying instead: _"(see `Exection model` for some caveats)"_ in order to keep things in one place?

rgommers (approved):
Overall LGTM, +1 for getting this in. Two minor comments to consider.

spec/purpose_and_scope.md
MarcoGorelli, line 128:
```suggestion
- Expectations on when the execution is happening (in an eager or lazy way) (see `execution model` for some caveats)
```

kkraus14 (approved):

MarcoGorelli:
thanks all, merging then

we can discuss propagation (or lack of) next time, but I'm glad we've been able to agree on this. it's something to be proud of. well done all! 🎉

PR #308, opened 2023-10-31
Add details to expectations for scalars

### LINKED ISSUE #305: What's the deal with Scalars?
Is the following pattern supported by the Standard?
```python
df: DataFrame
features = []
    if df.col(column_name).std() > 0:
        features.append(column_name)
return features
```
?

I think people are expecting that it should, but there's currently really no guarantee that it does.

All the API currently says about `Column.std` is that the return type should be a ducktyped float scalar. Let's try this:
```python
In [1]: class FancyFloatScalar:
   ...:     def __init__(self, value):
   ...:         self._value = value
   ...:     def __float__(self):
   ...:         return value
   ...:

In [2]: scalar = FancyFloatScalar(0.5)

In [3]: scalar > 0
TypeError                                 Traceback (most recent call last)
Cell In[3], line 1
----> 1 scalar > 0

TypeError: '>' not supported between instances of 'FancyFloatScalar' and 'int'

In [4]: scalar > 0.
TypeError                                 Traceback (most recent call last)
Cell In[4], line 1
----> 1 scalar > 0.

TypeError: '>' not supported between instances of 'FancyFloatScalar' and 'float'
```

What's the way of out of this?

cc @kkraus14 as you've said you wanted ducktyped scalars here

MarcoGorelli:
closes #305


MarcoGorelli:
I don't really have to strong an opinion on these. I can make a list of the methods which need to be supported, do you have preferences on the rest?

kkraus14:
> I don't really have to strong an opinion on these. I can make a list of the methods which need to be supported, do you have preferences on the rest?

I'm still generally -1 on ducktyping Python scalars because of the messiness that will come when doing things like `LIST` / `STRUCT` / `DECIMAL` types where I'd be in much stronger favor of having a `Scalar` class that can have a much smaller API area that implementations need to conform to.

MarcoGorelli:
agree - besides, the longer users stay within classes defined by the standard, the longer they know exactly what to expect

will update then, thanks

MarcoGorelli:
I've added a list of required methods

spec/design_topics/python_builtin_types.md
```diff
@@ -30,12 +30,70 @@ builtin types to CPU. In the above example, the `.mean()` call returns a
...
+- `__rfloordiv__`
+- `__truediv__`
+- `__rtruediv__`
+- `__neg__`
+- `__abs__`
+
+Furthermore, unless the library exclusively allows for lazy execution,
+it must also implement the following unary operations which return Python scalars:
+- `__int__`
+- `__float__`
```
MarcoGorelli, line 73: punt on these

MarcoGorelli:
I've updated to add a Scalar class, to note that `__bool__` may force compute (or raise, implementation-dependent), for now leaving out `__int__` / `__float__` / any other method to get a Python scalar out

MarcoGorelli:
I think everyone agreed with at least the content of this PR

There is further discussion to be had, but I think we can defer it until later, I suggest we start by merging the part which we agree on

spec/design_topics/python_builtin_types.md
```diff
@@ -18,7 +18,7 @@ class DataFrame:
         ...
 
 class Column:
-    def mean(self, skip_nulls: bool = True) -> float | NullType:
+    def mean(self, skip_nulls: bool = True) -> Scalar | NullType:
```
kkraus14, line 21: I think it's weird that this can return either a `Scalar` which is a protocol or a `NullType` which doesn't guarantee the same interface. It makes it hard to guarantee the ability to nicely method chain here.

MarcoGorelli: thanks, you're right - it should just be `-> Scalar`, correct? Because `Scalar` could be backed by a null value

spec/API_specification/dataframe_api/scalar_object.py
```diff
@@ -0,0 +1,88 @@
+from __future__ import annotations
+
+from typing import Any, Protocol
+
+__all__ = ['Scalar']
+
+
+class Scalar(Protocol):
```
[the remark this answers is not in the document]
MarcoGorelli: to be honest I'd be fine with departing completely from the idea of ducktyped Python scalars and just adding extra things (like `dtype` or `persist`) if they're useful

let's discuss this part in the next call

spec/API_specification/dataframe_api/typing.py
```diff
@@ -54,9 +48,6 @@
         Duration,
     ]
 
-# Type alias: Mypy needs Any, but for readability we need to make clear this
-# is a Python scalar (i.e., an instance of `bool`, `int`, `float`, `str`, etc.)
-Scalar = Any
 # null is a special object which represents a missing value.
 # It is not valid as a type.
 NullType = Any
```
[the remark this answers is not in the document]
MarcoGorelli: yes, brilliant point, thanks

I guess it could just return `None`?

spec/API_specification/dataframe_api/scalar_object.py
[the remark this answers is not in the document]
MarcoGorelli: Yes, I've added an example (`spec/API_specification/examples/05_scalars_example.py`)

Python scalars do in fact implement the Scalar protocol, so they can be passed without any issue

I think this is an argument against adding `dtype` and other methods to our scalars which aren't supported by Python ones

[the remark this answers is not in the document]
MarcoGorelli: > Python scalars do in fact implement the Scalar protocol, so they can be passed without any issue

at least, Python floats do

trying to do `fill_null('foo')` wouldn't be fine

I'm starting to see the writing on the wall for `FloatScalar` / `IntScalar` / `StringScalar` ...

spec/API_specification/dataframe_api/typing.py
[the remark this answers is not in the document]
MarcoGorelli: from https://github.com/data-apis/dataframe-api/pull/308#discussion_r1393308215 , I think we may need to end up with `NullScalar`

spec/design_topics/python_builtin_types.md
MarcoGorelli: if indeed we did go with https://github.com/data-apis/dataframe-api/pull/308#discussion_r1393308215, then `Scalar` could be a union of `FloatScalar`, `IntScalar`, `NullScalar`, ...

kkraus14: I don't know if we would want `Scalar` to be type erased similar to `Column`

spec/API_specification/dataframe_api/scalar_object.py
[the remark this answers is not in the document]
kkraus14: I imagine Scalars would be typed following the Column dtypes, which then allows us to define a consistent set of type handling and type promotion rules. Otherwise, if Scalars are not typed similarly to Columns, you may need to introspect the value of a Scalar for example in calling `fill_null` against an `Int8` dtype column with an `IntScalar` where the value is above what is supported by `Int8`.

spec/design_topics/python_builtin_types.md
MarcoGorelli: Could you clarify please? You could do `isinstance(value, namespace.FloatScalar)` to check the dtype

MarcoGorelli:
So, let's try to summarise what we're aiming for:
- some class which can store scalar values, so that the values can potentially stay lazy or reside on a different device
- if a function accepts an integer, it should be possible to pass either a `Scalar` or a Python scalar
- (Keith): be able to tell the dtype of a scalar

The current suggestion in this PR isn't actually sufficient, because it defines `__bool__`. However, Python `str` scalars don't implement many of the listed methods, such as `__abs__`, `__divmod__`, and many more...

If we really want interchangeability with Python scalars:
- define `NumericScalar`, `BoolScalar`, `StringScalar`, `DateScalar`, `DatetimeScalar`, `DurationScalar`, `NullScalar` setting out expectations for what each is expected to implemented, and which trigger execution. The annoying thing is that then you have to implement all these classes, and that you don't even necessarily know which to use (e.g. in `get_value` the return type could be anything)

If we don't need interchangeability with Python scalars:
- Keep `Scalar` as a class, but in some methods, the argument can be something like `AnyScalar`, where `AnyScalar = Union[Scalar, PythonScalar` and `PythonScalar = Union[int, float, str]`
- `Scalar` can then have some extra attributes such as `.dtype` and `.persist`
- note that some methods may raise depending on the underlying scalar (for example, `__abs__` would raise if a string scalar)

spec/API_specification/dataframe_api/typing.py
```diff
@@ -169,15 +167,31 @@ def __column_consortium_standard__(
...
 
 
+PythonScalar = Union[str, int, float, bool]
+BoolScalar = Union[bool, Scalar]
+FloatScalar = Union[float, Scalar]
+IntScalar = Union[int, Scalar]
+NumericScalar = Union[FloatScalar, IntScalar]
+StringScalar = Union[str, Scalar]
+AnyScalar = Union[PythonScalar, Scalar]
+NullType = Namespace.NullType
```
kkraus14, line 177: Is this a null scalar singleton or a null dtype or something else? If it's a null scalar singleton then I think it needs to be added to bunch of these `Union` types.

spec/API_specification/dataframe_api/typing.py
```diff
@@ -169,15 +167,31 @@ def __column_consortium_standard__(
         ...
 
 
+PythonScalar = Union[str, int, float, bool]
+BoolScalar = Union[bool, Scalar]
+FloatScalar = Union[float, Scalar]
+IntScalar = Union[int, Scalar]
```
kkraus14, line 173: Without `Int8Scalar`, `Int16Scalar`, `Int32Scalar`, I think we need to do value introspection to figure out how to handle type promotion or throwing or whatnot.

Should we have a Scalar class for each dtype?

spec/design_topics/python_builtin_types.md
kkraus14: Apologies, we're aligned here. We have different classes per type as opposed to just a top level `Scalar` class where type specific functionality is hidden underneath the class (similar to our `Column` class).

spec/API_specification/dataframe_api/typing.py
MarcoGorelli: These are just type hints

The idea is:
- Scalar is the Standard class for scalars. It has `dtype` and `persist` methods

FloatScalar is just for type hint purposes and means "either a Python float or a standard Scalar

I could simplify and just have PythonScalar and Scaler, this is becoming too complex

MarcoGorelli:
Have updated. We now have:
- `Scalar`: class backed by a scalar, which allows data to reside on GPU or stay lazy. Has `dtype` property and `persist` method
- `AnyScalar`: type hint meaning "either a Python scalar, or a Scalar`. This makes explicit that both
  ```python
  df: DataFrame
  df > 3  # supported
  df > df.col('a').mean()  # also supported
  ```
  are supported (`3` is a Python scalar, `df.col('a').mean()` is a `Scalar`)

`null` is just a singleton null value. It is not a `Scalar`, as it does not have a `dtype` defined. So, for `fill_nan`, the fill value can be `float | NullType | Scalar`, meaning you can do:
```python
df: DataFrame
ns = df.__dataframe_namespace__()
df.col('a').fill_nan(3.)  # supported (`float`)
df.col('a').fill_nan(ns.null)  # also supported (`NullType`)
df.col('a').fill_nan(df.col('b').mean())  # also supported (`Scalar`)
```

From the discussion in https://github.com/data-apis/dataframe-api/pull/294, I've also added `Scalar.persist`, which seems very much needed.
If people are OK with this design, then I think #294 will finally be ready (sorry @cbourjau it's taken so long, and thanks for all the discussion which your example has spurred!)

spec/API_specification/dataframe_api/typing.py
```diff
@@ -169,15 +167,31 @@ def __column_consortium_standard__(
         ...
 
 
+PythonScalar = Union[str, int, float, bool]
```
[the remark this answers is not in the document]
MarcoGorelli: there's already `namespace.date` to create a scalar which can be used for date columns (e.g. see how it's used in TPCH Q1:
```python
    mask = lineitem.col("l_shipdate") <= namespace.date(1998, 9, 2)
```
), I think that's enough?

kkraus14 (approved):

MarcoGorelli:
thanks for your review! leaving open a bit, if others have objections please speak up - else, really excited to be moving forwards, so glad we've found common ground on the design here

PR #310, opened 2023-10-31
Add note about parent dataframes and cross-dataframe column comparisons

MarcoGorelli:
I think we can relax many of these restrictions for the case of free-standing columns (i.e. those without a parent dataframe, created via `column_from_1d_array`) - just working through it to check, will update this week

### DIFF
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -238,6 +238,11 @@ class Column(Protocol):
         Returns
         -------
         Column
+
+        Notes
+        -----
+        `other`'s parent DataFrame must be the same as `self`'s - else,
+        the operation is unsupported and may vary across implementations.
         """
         ...
 
@@ -257,6 +262,11 @@ class Column(Protocol):
         Returns
         -------
         Column
+
+        Notes
+        -----
+        `other`'s parent DataFrame must be the same as `self`'s - else,
+        the operation is unsupported and may vary across implementations.
         """
         ...
 
@@ -274,6 +284,11 @@ class Column(Protocol):
         Returns
         -------
         Column
+
+        Notes
+        -----
+        `other`'s parent DataFrame must be the same as `self`'s - else,
+        the operation is unsupported and may vary across implementations.
         """
         ...
 
@@ -291,6 +306,11 @@ class Column(Protocol):
         Returns
         -------
         Column
+
+        Notes
+        -----
+        `other`'s parent DataFrame must be the same as `self`'s - else,
+        the operation is unsupported and may vary across implementations.
         """
         ...
 
@@ -308,6 +328,11 @@ class Column(Protocol):
         Returns
         -------
         Column
+
+        Notes
+        -----
+        `other`'s parent DataFrame must be the same as `self`'s - else,
+        the operation is unsupported and may vary across implementations.
         """
         ...
 
@@ -325,6 +350,11 @@ class Column(Protocol):
         Returns
         -------
         Column
+
+        Notes
+        -----
+        `other`'s parent DataFrame must be the same as `self`'s - else,
+        the operation is unsupported and may vary across implementations.
         """
         ...
 
@@ -342,6 +372,11 @@ class Column(Protocol):
         Returns
         -------
         Column
+
+        Notes
+        -----
+        `other`'s parent DataFrame must be the same as `self`'s - else,
+        the operation is unsupported and may vary across implementations.
 
         Raises
         ------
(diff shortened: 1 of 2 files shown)


spec/API_specification/dataframe_api/column_object.py
```diff
@@ -17,8 +18,43 @@ class Column(Protocol):
...
+    - If two columns were retrieved from different dataframes,
+      then there is no guarantee about how or whether they can be combined and
+      compared, this may vary across implementations.
+    - If two columns are both "free-standing" (i.e. not retrieved from a dataframe
+      but constructed directly from a 1D array or sequence), then they can be
+      combined and compared with each other. Note, however, that they still can't
+      be compared or combined with columns retrieved from a dataframe.
     """
+    @property
+    def dataframe(self) -> DataFrame | None:
```
kkraus14, line 37: Do we want this to be part of the public API? Do we have any ideas where a user of this API would want to retrieve the owning DataFrame as opposed to the library just using the information internally?

MarcoGorelli: I don't think it needs to be part of the public API, no, I just thought this is what you were suggesting on the call - I may have misunderstood though

MarcoGorelli: @kkraus14 gentle ping on either this or #298

spec/API_specification/dataframe_api/column_object.py
```diff
@@ -17,9 +19,44 @@ class Column(Protocol):
...
+
+    - If two columns were retrieved from the same dataframe,
+      then they can be combined and compared at will.
+    - If two columns were retrieved from different dataframes,
+      then there is no guarantee about how or whether they can be combined and
+      compared, this may vary across implementations.
+    - If two columns are both "free-standing" (i.e. not retrieved from a dataframe
+      but constructed directly from a 1D array or sequence), then they can be
+      combined and compared with each other. Note, however, that they still can't
+      be compared or combined with columns retrieved from a dataframe.
```
kkraus14, line 35: Instead of "they still can't be compared or combined with columns retrieved from a dataframe" I think we can have this also be "there is no guarantee about how or whether they can be compared or combined with columns retrieved from a dataframe, this may vary across implementations"

spec/API_specification/dataframe_api/column_object.py
```diff
@@ -17,9 +19,44 @@ class Column(Protocol):
...
     """
 
+    @property
+    def dataframe(self) -> DataFrame | None:
```
kkraus14, line 39: What are your thoughts on making this `parent_dataframe` or `owning_dataframe`. Having it just be `.dataframe` is a bit ambiguous without looking at the docs and my first interpretation would be that it returns a DataFrame with this column as the single column.

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -172,8 +177,8 @@ def filter(self, mask: Column) -> Self:
 
         Notes
         -----
-        Some participants preferred a weaker type Arraylike[bool] for mask,
-        where 'Arraylike' denotes an object adhering to the Array API standard.
+        `mask`'s parent DataFrame must be `self` - else,
```
kkraus14, line 180: I think should be "must be the same as `self`'s" instead of "must be `self`"

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -136,6 +136,11 @@ def get_rows(self, indices: Column) -> Self:
         Returns
         -------
         DataFrame
+
+        Notes
+        -----
+        `indices`'s parent DataFrame must be `self` - else,
```
kkraus14, line 142: +1

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -201,6 +206,11 @@ def assign(self, *columns: Column) -> Self:
         Returns
         -------
         DataFrame
+
+        Notes
+        -----
+        All of `columns`'s parent DataFrame must be `self` - else,
```
kkraus14, line 212: +1

spec/API_specification/dataframe_api/column_object.py
kkraus14: I think if we're going to potentially throw this should be part of the public API to allow a user to introspect what's going on. Going to resolve this conversation.

spec/API_specification/dataframe_api/dataframe_object.py
MarcoGorelli: `self` is itself a `DataFrame` though, so I think that `mask.parent_dataframe` should indeed be `self` here?

spec/API_specification/dataframe_api/column_object.py
MarcoGorelli: yup, good call

kkraus14 (approved):

MarcoGorelli:
Thanks for your careful review, much appreciated 🙏 

Leaving open a bit longer in case others have objections, then will merge

PR #317, opened 2023-11-10
add column.shift

MarcoGorelli:
Noticed in scikit-lego, but really any library related to forecasting would need this


spec/API_specification/dataframe_api/column_object.py
```diff
@@ -811,6 +811,23 @@ def rename(self, name: str) -> Self:
         """
         ...
 
+    def shift(self, offset: int) -> Self:
```
kkraus14, line 814: Dependent on #308, but offset should ideally be of type `Scalar` instead of `int`

kkraus14 (approved):

MarcoGorelli:
agree, let's fix it up as part of that

PR #322, opened 2023-11-16
improve persist example

MarcoGorelli:
we can call `persist` after `.std()` and `__gt__`

### DIFF
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -982,7 +982,8 @@ class DataFrame(Protocol):
                 result = df.std() > 0
                 result = result.persist()
                 for column_name in df.column_names:
-                    if result.col(column_name).get_value(0) > 0:
+                    include_column = result.col(column_name).get_value(0).persist()
+                    if include_column:
                         features.append(column_name)
 
             instead of this:
--- a/spec/API_specification/dataframe_api/dataframe_object.py
+++ b/spec/API_specification/dataframe_api/dataframe_object.py
@@ -982,8 +982,7 @@ class DataFrame(Protocol):
                 result = df.std() > 0
                 result = result.persist()
                 for column_name in df.column_names:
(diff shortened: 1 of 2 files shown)


spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -991,7 +992,8 @@ def persist(self) -> Self:
                 df: DataFrame
                 features = []
                 for column_name in df.column_names:
-                    # Do NOT do this!
+                    # Do NOT call `persist` on a `DataFrame` within a for-loop!
+                    # This may re-triggers the same computation multiple times
```
kkraus14, line 996:
```suggestion
                    # This may re-trigger the same computation multiple times
```

kkraus14 (approved):

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -978,10 +978,12 @@ def persist(self) -> Self:
             .. code-block:: python
 
                 df: DataFrame
-                df = df.persist()
                 features = []
+                result = df.std() > 0
+                result = result.persist()
                 for column_name in df.column_names:
-                    if df.col(column_name).std() > 0:
+                    include_column = result.col(column_name).get_value(0).persist()
```
MarcoGorelli, line 985: Updating to use `Scalar.persist` as I still feel really uneasy about `__bool__` automatically triggering compute

I'd quite like a simple rule like "to be able to bring a scalar into Python, you need to call `persist` on that scalar first". This is much simpler than defining "persistedness" propagation for dataframes, and addresses my concern about people accidentally having compute triggered for them because they don't know that `if scalar:` calls `__bool__`

MarcoGorelli: Maybe this one's OK actually

If you've done `df.persist`, then `df.get_value(...)` and `df.col(...)` shouldn't need re-persisting

PR #325, opened 2023-11-17
remove dtype from to_array

MarcoGorelli:
similar to https://github.com/data-apis/dataframe-api/pull/293


spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -903,26 +903,26 @@ def drop_nulls(
...
+        - Int64() -> 'int64'
+        - UInt8() -> 'uint8'
+        - UInt16() -> 'uint16'
+        - UInt32() -> 'uint32'
+        - UInt64() -> 'uint64'
+        - Float32() -> 'float32'
+        - Float64() -> 'float64'
+
+        and multiple columns' dtypes are combined according to the Array API's
+        type promotion rules.
```
kkraus14, line 925: Maybe link off to the type promotion rules? https://data-apis.org/array-api/latest/API_specification/type_promotion.html

Also, they don't define rules for mixed integer and floating-point types and specify it as implementation defined. Are we good with that as well?

MarcoGorelli: good call, have added

yeah that sounds fine to me, you OK with it too?

spec/API_specification/dataframe_api/dataframe_object.py
```diff
@@ -813,26 +813,26 @@ def drop_nulls(
...
```
kkraus14, line 835: Did we want to link to the array API docs on these type promotion rules?

kkraus14 (approved):

spec/API_specification/dataframe_api/dataframe_object.py
MarcoGorelli: sorry, forgot to push, have done so now

PR #326, opened 2023-11-20
Add DataFrame.iter_columns() and simplify

MarcoGorelli:
This simplifies a few examples and makes things more ergonomic

Can't just use `DataFrame.columns` as pandas / Polars already use that for column names unfortunately


cbourjau:
> Can't just use DataFrame.columns as pandas / Polars already use that for column names unfortunately

Out of curiosity: Is it a general design principle to avoid function names that exist in Pandas/Polars but have different semantics there?

MarcoGorelli:
approved in today's call

PR #327, opened 2023-11-20
Remove schema from dataframe_from_2d_array, use names instead

MarcoGorelli:


kkraus14 (approved):

MarcoGorelli:
thanks - we'll probably have to add `Column.cast`, but that'll be separate

PR #328, opened 2023-11-20
Consistently use ns for namespace in docs/examples

MarcoGorelli:
`namespace = dataframe.__dataframe_namespace__()` is quite verbose, how about just using a 2-letter acronym in the examples? most people alias namespace to 2-letter acronyms anyway

### DIFF
--- a/spec/API_specification/dataframe_api/__init__.py
+++ b/spec/API_specification/dataframe_api/__init__.py
@@ -293,10 +293,10 @@ def date(year: int, month: int, day: int) -> Scalar:
     Examples
     --------
     >>> df: DataFrame
-    >>> ns = df.__dataframe_namespace__()
+    >>> pdx = df.__dataframe_namespace__()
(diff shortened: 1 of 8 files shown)


rgommers:
Thanks Marco. This makes sense to me. I was never quite a fan of the `namespace` name. `ns` seems better - or something like `pdx` would work too (we use `xp` on the array side).

MarcoGorelli:
thanks - yeah thought of `pdx` too but it struck me as too `pandas-centric`

rgommers:
Makes sense. Let's see if anyone else has a strong opinion - if not, then let's merge within a couple of days.

kkraus14:
For what it's worth, I interpret `pdx` as `Python DataFrame X` and unrelated to pandas, so I'd be +1 to that. `ns` makes me think nanoseconds 😆

kkraus14 (approved):

PR #330, opened 2023-11-21
fixup typing of Duration and Datetime dtypes

MarcoGorelli:
add the `time_unit` and `time_zone` attributes


MarcoGorelli:
merging as this is really minor and blocks a pr of mine

PR #331, opened 2023-11-21
add Column.len, which returns Scalar, instead of Column.__len__ (returning int)

MarcoGorelli:


kkraus14:
@shwina what is your thoughts on this one? I know cuDF is eager and so it always has the length of columns accessible, but given its scalar class is built to be on device (ignoring the internal implementation of how it handles both host and device access), do you have any thoughts here?

shwina:
So I expect the length to be used for things like comparisons or equality checking which does work between Scalars and plain Python integers in cuDF. This should be an OK change

MarcoGorelli:
approved in today's call

PR #335, opened 2023-11-23
add release workflow for dataframe api

MarcoGorelli:
As discussed in the last call, let's start publishing the dataframe-api package to PyPI

This means consumers will be able to type-check their usage of the dataframe api, and this will tell them when they're staying within the standard and when they're not

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -1,0 +1,36 @@
+name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+
+on: push
+
+jobs:
+  build-n-publish:
+    name: Build and publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: "3.x"
+
+    - name: Install pypa/build
+      run: >-
+        python3 -m
+        pip install
+        build
+        --user
+    - name: Build a binary wheel and a source tarball
+      run: >-
+        python3 -m
+        build
+        --sdist
+        --wheel
+        --outdir dist/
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,36 @@
...
+        build
+        --user
+    - name: Build a binary wheel and a source tarball
+      run: >-
+        python3 -m
+        build
+        --sdist
+        --wheel
+        --outdir dist/
+        spec/API_specification
```
rgommers, line 30: minor: these would be much easier to read as one-liners. E.g.:
```
run: |
  python3 -m pip install build
```

.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,36 @@
...
+      uses: actions/setup-python@v4
+      with:
+        python-version: "3.x"
+
+    - name: Install pypa/build
+      run: >-
+        python3 -m
+        pip install
+        build
+        --user
```
rgommers, line 22: minor: `--user` is basically obsolete, and should not be needed in CI; you can remove it here I'd say.

.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,36 @@
+name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+
+on: push
```
rgommers, line 3: The conditions don't seem quite right here. It only runs on `push`, not on `pull`. And then the publish stage only publishes on any tags. Should this be something like:
```yaml
on:
  pull_request:
    branches:
      - main
  push:
    tags:
      - 'v*'
``` 
?

rgommers (commented):
Thanks Marco. This looks pretty good to me - a few comments. The deploy token exists I see, so this looks about good to go.

MarcoGorelli:
closing just to keep my open PRs down, but others are welcome to take this forward

PR #338, opened 2023-12-07
Add cast

### LINKED ISSUE #203: Methods for casting from one dtype to another
Should the standard support casting columns from one dtype to another? Like, Int64 -> Float64?

Could have:
- `DataFrame.cast_columns`, which takes a mapping between column names and dtypes
- `Column.cast` which accepts a single dtype

MarcoGorelli:
closes #203

### DIFF
--- a/spec/API_specification/dataframe_api/column_object.py
+++ b/spec/API_specification/dataframe_api/column_object.py
@@ -1061,7 +1061,9 @@ class Column(Protocol):
     def cast(self, dtype: DType) -> Self:
         """Cast specified columns to specified dtypes.
 
-        Cross-kind casting (e.g. integer to string) is not supported and may
-        vary across implementations.
+        The following is not specified and may vary across implementations:
+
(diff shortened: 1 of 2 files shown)


spec/API_specification/dataframe_api/column_object.py
```diff
@@ -1057,3 +1057,11 @@ def persist(self) -> Self:
             at most once per dataframe, and as late as possible in the pipeline.
         """
         ...
+
+    def cast(self, dtype: DType) -> Self:
+        """Cast specified columns to specified dtypes.
+
+        Cross-kind casting (e.g. integer to string) is not supported and may
+        vary across implementations.
```
kkraus14, line 1065: I think we need to specify what is vs isn't supported here? I.E. is casting `int64` to `int32` supported? What's the behavior if a value doesn't fit in `int32` space?

MarcoGorelli:
from the call:
- overflows are implementation-specific
- cross-kind casting is implementation-specific
- the rest should be supported

Issue #279, opened 2023-10-12
Add dunder method for Arrow C Data Interface to DataFrame and Column objects


jorisvandenbossche:
The Python Arrow community is adding a public way to interchange data through the C Data Interface, using PyCapsule objects holding the C struct, similary as DLPack's python interface: http://crossbow.voltrondata.com/pr_docs/37797/format/CDataInterface/PyCapsuleInterface.html

We have DLPack support at the Buffer level, and similarly, I think it would be useful to add Arrow support at the DataFrame and Column level. 

Concretely, I would propose adding an optional `__arrow_c_schema__`, `__arrow_c_array__` and `__arrow_c_stream__` methods to both the `DataFrame` and `Column` interchange objects. Those methods would be optional, with their presence indicating that this specific implementation of the interchange object supports the Arrow interface.  
Consumers of the interchange protocol could then check for the presence of those methods, and try them first for an easier and faster conversion, and otherwise use the standard APIs through the Column and Buffer objects (example: pyarrow and polars interchanging data).

It might be a bit strange to add both the array and stream interface methods, but that is due to that the interchange protocol hasn't really made a distinction between a single chunk vs a chunked object. But I think the array method could then raise an error if the DataFrame or Column still exists of more than 1 chunk.

kkraus14:
I think we'd want to use the C Device Data Interface as opposed to the C Data Interface in order to support non-CPU memory as well?

kkraus14:
I think we should just wait until there's a Python protocol for the C Device Data Interface then instead of plumbing in the C Data Interface. That gives a single implementation that could be universally supported as opposed to adding something that downstream consumers will either have to check for existence or check if it throws or something else.

rgommers:
I see the device data interface is still marked as experimental in Arrow (https://arrow.apache.org/docs/dev/format/CDeviceDataInterface.html). Is there a timeline for it appearing in PyArrow?

jorisvandenbossche:
> Is there a timeline for it appearing in PyArrow?

(for the device interface) Hopefully in the next release (15.0), depending on the progress in https://github.com/apache/arrow/issues/38325)

---

Some questions here on what the expected return data would be when the data does not match default Arrow types exactly. 
 
I think in general DLPack is expected to be zero-copy (and if that is not possible (because of the data type, the device, or being distributed, ...), you raise an error instead). The question is whether we want to define the same expectation here? 

[...]

PR #342, opened 2023-12-21
Add methods for the Arrow PyCapsule Protocol to DataFrame/Column interchange protocol objects

jorisvandenbossche:
Addresses https://github.com/data-apis/dataframe-api/issues/279

Currently, I added schema and array support to `Column`, and schema and stream support to `DataFrame`. I _think_ that's the typical way those objects are used (chunked dataframe, gets chunks+columns, and then single-chunk column), but for completeness (and because the spec doesn't distinguish between single vs multi-chunk objects) we can probably add both the array and stream method to both the DataFrame and Column object. 

I also added the schema method to both DataFrame and Column, although those are strictly speaking not "schema" objects. But we don't have a dedicated schema object in the interchange spec (DataFrame has no attribute, and the Column.dtype attribute is a plain tuple)

### DIFF
--- a/protocol/dataframe_protocol.py
+++ b/protocol/dataframe_protocol.py
@@ -350,6 +350,41 @@ class Column(ABC):
         """
         pass
 
+    def __arrow_c_schema__(self) -> object:
+        """
+        Export the data type of the Column to a Arrow C schema PyCapsule.
+
+        Returns
+        -------
+        PyCapsule
+        """
+        pass
+
+    def __arrow_c_array__(
+        self, requested_schema: Optional[object] = None
+    ) -> Tuple[object, object]:
+        """
+        Export the Column as an Arrow C array and schema PyCapsule.
+
+        If the Column consists of multiple chunks, this method should raise
+        an error.
+
+        Parameters
+        ----------
+        requested_schema : PyCapsule, default None
+            The schema to which the dataframe should be casted, passed as a
+            PyCapsule containing a C ArrowSchema representation of the
+            requested schema.
+            If None, the array will be returned as-is, with a type matching the
+            one returned by ``__arrow_c_schema__()``.
+
+        Returns
+        -------
+        Tuple[PyCapsule, PyCapsule]
+            A pair of PyCapsules containing a C ArrowSchema and ArrowArray,
+            respectively.
+        """
+        pass
 
 #    def get_children(self) -> Iterable[Column]:
 #        """
@@ -488,5 +523,34 @@ class DataFrame(ABC):
 
         Note that the producer must ensure that all columns are chunked the
         same way.
+        """
+        pass
+
+    def __arrow_c_schema__(self) -> object:
+        """
+        Export the schema of the DataFrae to a Arrow C schema PyCapsule.
+
+        Returns
+        -------
+        PyCapsule
+        """
+        pass
+
+    def __arrow_c_stream__(self, requested_schema: Optional[object] = None) -> object:
+        """
+        Export the DataFrame as an Arrow C stream PyCapsule.
+
+        Parameters
+        ----------
+        requested_schema : PyCapsule, default None
+            The schema to which the dataframe should be casted, passed as a
+            PyCapsule containing a C ArrowSchema representation of the
+            requested schema.
+            If None, the array will be returned as-is, with a type matching the
+            one returned by ``__arrow_c_schema__()``.
+
+        Returns
+        -------
+        PyCapsule
         """
         pass


kkraus14:
What should the behavior be if the library has its data on say a GPU instead of CPU?

Throw? Copy to CPU?

If we support this CPU only protocol why don't we support other CPU only protocols as well?

jorisvandenbossche:
> What should the behavior be if the library has its data on say a GPU instead of CPU? Throw? Copy to CPU?

That's one thing I still needed to add: if you do not support this protocol, I would propose to indicate this by not having the methods (so you get do `hasattr(obj, "..")` as a proxy for checking if the object supports this protocol). 

So if we decide that GPU libraries shouldn't support this, then they should simply not add those methods.

Personally I don't have much experience with GPU, but based on following earlier discussions, I think the general preference is that there is no silent copy?

Certainly given the fact we want to expand this protocol to support GPU data as well (https://github.com/apache/arrow/issues/38325), I think not implementing this right now is the best option for objects backed by GPU data.

> If we support this CPU only protocol why don't we support other CPU only protocols as well?

Which other CPU-only protocol are you thinking about? I am not aware of any other dataframe protocol. 
(there are of course _array_ protocols, which you might be hinting on, but we already do have DLPack on the buffer/array level that already supports devices. So adding another array protocol at that level is much less needed given that most array libraries nowadays support DLPack. Personally I think adding general Python buffer protocol support would be nice, though)

MarcoGorelli (approved):
Looks good!

If these methods aren't strictly required, maybe this could be noted in the docstring? I think that would address Keith's concern

Sorry for having taking a while to respond here, it's taken me a while to fill gaps in my knowledge (thanks Joris for patiently explaining and answering questions!)

rgommers:
> So if we decide that GPU libraries shouldn't support this, then they should simply not add those methods.

I agree - if they'd start copying data to host memory now to support the CPU-only protocol, that would make it much harder to later add the device-agnostic version.

If there was movement and a timeline on https://github.com/apache/arrow/issues/38325, it'd be easier to say "let's not do the CPU-only version". However, it looks like that may take quite a while. Somehow work on these protocols is always slow going (DLPack was/is no exception). So it seems reasonable to me to start supporting the CPU protocol now.

There is indeed the fragmentation on the array side, with too many different CPU-only protocols. However, on the dataframe side things are cleaner. If things could stay limited to:

- `__dataframe__` and `from_dataframe` at the Python API level
- the Arrow CPU-only protocol as dunder methods
- the Arrow device-agnostic protocol as dunder methods

then it looks like things are in pretty good shape.

rgommers (commented):
Thanks @jorisvandenbossche! I don't have practical experience with this protocol, but it makes sense to me conceptually to add, and this PR looks like a good start to me. It's pretty self-contained, which is nice.

I have two main questions:

(1) Should these methods say to only implement them if the data is natively stored in Arrow format under the hood? If not, it may turn into yet more work for implementers - and require copying on the producer side.

(2) Exporting `PyCapsule` objects always comes with the "who owns this" question. For DLPack we worked hard at eliminating that problem by specifying a public function for it and making clear that the `PyCapsule` was supposed to be consumed once after creating and then immediately discarded, so it wasn't possible for `PyCapsule`'s to hang around in user land. I think the intent here is similar, but it's not mentioned at all. Should there be a requirement on this?

protocol/dataframe_protocol.py
```diff
@@ -490,3 +525,32 @@ def get_chunks(self, n_chunks: Optional[int] = None) -> Iterable["DataFrame"]:
...
+        """
+        Export the schema of the DataFrae to a Arrow C schema PyCapsule.
+
+        Returns
+        -------
+        PyCapsule
+        """
+        pass
+
+    def __arrow_c_stream__(self, requested_schema: Optional[object] = None) -> object:
```
kkraus14, line 539: How does one differentiate between a DataFrame and a struct column here (assuming those will be supported in the future)?

kkraus14:
I am still a strong -1 on this proposal.

The array API didn't implement `__array__` or `__array_interface__` despite being more widely adopted and used than the equivalent Arrow protocols being added here. It wasn't implemented because they didn't want to fragment implementations or cause downstream pain in leveraging non-CPU, non-numpy implementations.

We have already defined a new interchange protocol that can handle Arrow formatted data zero copy in addition to handling non-CPU memory and non-arrow memory layouts.

MarcoGorelli:
That's OK

@jorisvandenbossche we can still use this in pandas/polars/pyarrow `from_dataframe`:

```python
def from_dataframe(df: SupportsInterchange, *, allow_copy: bool = True) -> DataFrame:
    if hasattr(df, '__arrow_c_schema__'):
        return fastpath_constructor(df, allow_copy=allow_copy)
    return _from_dataframe(
        df.__dataframe__(allow_copy=allow_copy),  # type: ignore[arg-type]
        allow_copy=allow_copy,
    )
```
even if it's not officially part of the interchange protocol

jorisvandenbossche:
> The array API didn't implement `__array__` or `__array_interface__` despite being more widely adopted and used than the equivalent Arrow protocols being added here. It wasn't implemented because they didn't want to fragment implementations or cause downstream pain in leveraging non-CPU, non-numpy implementations.

I understand why the array API didn't implement those, but not sure how that is relevant for this discussion? I am not proposing to add this protocol based on some popularity metric, but because it is adding missing features and convenience. The array protocols you mentioned also wouldn't add much additional value for the Array API because it already had DLPack covering the same. 
For the DataFrame interchange protocol, though, that is not the case. What I am proposing here has additional value on top of what we already have (allow accessing the data at DataFrame/Column level instead of only at Buffer level + exposing the data through a stable C ABI interface, similarly as DLPack does for the arrays)

@kkraus14 what is not clear to me, is your objection mainly based on the lack of non-cpu device support, or more in general? Assume for a moment that the Arrow community already resolved the discussion around adding device support for the Arrow PyCapsule protocol, and we could directly add it here as well. Would you then be OK with it?

> we can still use this in pandas/polars/pyarrow interchange.from_dataframe,  ... even if it's not officially part of the interchange protocol

@MarcoGorelli that's certainly possible and would already improve that part of the interchange's usecases, but that still doesn't enable usage of the interchange protocol in places that require Arrow memory (using duckdb as the typical example, they support Arrow memory, and doubt they would ever support the interchange protocol directly. They could of course already easily support it if they want by depending on pyarrow, but that dependency isn't needed with this PR)

kkraus14:
> @kkraus14 what is not clear to me, is your objection mainly based on the lack of non-cpu device support, or more in general? Assume for a moment that the Arrow community already resolved the discussion around adding device support for the Arrow PyCapsule protocol, and we could directly add it here as well. Would you then be OK with it?

I would be more okay with it, but I still feel funny about giving Arrow memory layout arguably preferential treatment over non Arrow memory layout, though we already do that for strings more or less.

I.E. if someone uses a sentinel value or a *byte* mask for representing nulls, should they not support this protocol or should they support it via a copy + conversion? Both are unideal from the perspective of a library not using Arrow format for representing nulls supporting a standard.

MarcoGorelli:
Note that cuDF currently works perfectly fine with the PyCapsule Interface, whereas its interchange protocol fails with the most basic of test cases https://github.com/rapidsai/cudf/issues/17282#issuecomment-2464826727

I'll also note that Vegafusion has replaced its usage of the dataframe interchange protocol with the PyCapsule Interface

PR #349, opened 2024-02-13
Add `slice_rows` to interchange protocol

### LINKED ISSUE #204: How to slice rows? Can it fit into the interchange, or is the standard required?
Came across this yesterday: https://github.com/microsoft/vscode-jupyter/pull/13951
```python
        elif _VSCODE_builtins.hasattr(df, "to_pandas"):
            df = df.to_pandas().iloc[start:end]
```

This could be improved if they only needed to convert to pandas the part of the data which they needed, i.e.:
```python
elif hasattr(df, "__dataframe__"):
    df = pd.api.interchange.from_dataframe(df.__dataframe__().slice_rows(start, end))
```

Or, if the Standard were actually available:
```python
elif hasattr(df, "__dataframe__") and hasattr(df, "__dataframe_standard__"):
    df = pd.api.interchange.from_dataframe(df.__dataframe_standard().slice_rows(start, end).dataframe)
```

So, could this fit into the interchange protocol, or would filtering by rows only be possible for libraries implementing the standard?

It just seems a bit much to require the standard (especially if a separate compatibility package is required for it) if all people need to do is slice rows

MarcoGorelli:
closes #204

### DIFF
--- a/protocol/dataframe_protocol.py
+++ b/protocol/dataframe_protocol.py
@@ -482,7 +482,7 @@ class DataFrame(ABC):
     def slice_rows(
         self,
         offset: int,
-        length: int | None,
+        size: int | None,
     ) -> DataFrame:
         """
         Select a subset of rows corresponding to a slice.
@@ -491,8 +491,8 @@ class DataFrame(ABC):
         ----------
         offset : int
             Row offset to start slice at.
-        length : int or None
-            Length of slice. If None, slice to end of DataFrame.
+        size : int or None
+            Number of rows of the resulting slice. If None, slice to end of DataFrame.
 
         Returns
         -------
--- a/protocol/dataframe_protocol.py
+++ b/protocol/dataframe_protocol.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from abc import (
     ABC,
     abstractmethod,
@@ -473,6 +475,31 @@ class DataFrame(ABC):
     def select_columns_by_name(self, names: Sequence[str]) -> "DataFrame":
         """
         Create a new DataFrame by selecting a subset of columns by name.
+        """
+        pass
+
+    @abstractmethod
+    def slice_rows(
+        self,
+        start: int | None = None,
+        stop: int | None = None,
+        step: int | None = None,
+    ) -> DataFrame:
+        """
+        Select a subset of rows corresponding to a slice.
+
+        Parameters
+        ----------
+        start : int or None
+            If None, starts at 0.
+        stop : int or None
+            If None, stops at the end.
+        step : int or None
+            If None, uses the default step of 1.
+
+        Returns
+        -------
+        DataFrame
         """
         pass
 
--- a/protocol/dataframe_protocol.py
+++ b/protocol/dataframe_protocol.py
@@ -481,21 +481,18 @@ class DataFrame(ABC):
     @abstractmethod
     def slice_rows(
         self,
-        start: int | None = None,
-        stop: int | None = None,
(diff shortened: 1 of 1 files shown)


kkraus14:
I think this current implementation leaves a lot to be desired. I.E. in the case of something like pandas or other dataframe library that doesn't use the Arrow memory layout under the hood, they'd presumably materialize arrow on the `__dataframe__` call and then have to slice the arrow format memory, which if containing strings or has a step size, isn't free. This is already potentially a problem in selecting columns as well, so I guess this inefficiency is nothing new?

Additionally, it makes it a bit hard to reason about when the producer vs when the consumer should do row selection. I.E. if Polars is consuming data from say PyArrow, I imagine Polars would rather handle row slicing itself (assuming you'll hit a situation where it's not pure pointer arithmetic). Now in the situation of Pandas consuming data from say Polars, you'd probably want Polars to handle the row slicing.

Arrow interchange protocols handle the slicing case (ignoring step size) by allowing specifying an offset and a size. Maybe we can do something similar here?

MarcoGorelli:
sounds good, thanks

protocol/dataframe_protocol.py
```diff
@@ -476,6 +478,28 @@ def select_columns_by_name(self, names: Sequence[str]) -> "DataFrame":
         """
         pass
 
+    @abstractmethod
+    def slice_rows(
+        self,
+        offset: int,
+        length: int | None,
```
kkraus14, line 485: Should we call this `size` to make it consistent with the property we have on the `Column` objects? https://github.com/data-apis/dataframe-api/blob/main/protocol/dataframe_protocol.py#L203

kkraus14:
Do we expect / want to encourage developers using dataframe libraries to explicitly call `__dataframe__` themselves as opposed to using `libraryx.from_dataframe(...)`? It feels a bit funky to me currently that we go from say:
```
pl_df = ...  # My polars dataframe
pdf = pandas.from_dataFrame(pl_df)
```

to:
```
pl_df = ...  # My polars dataframe
pdf = pandas.from_dataframe(pl_df.__dataframe__().select_columns(...).slice_rows(...))
```

My 2c is that this is just highlighting the lack of standard API here and that the experience should be something along the lines of (ignoring API names for column selection and row slicing):
```
pl_df = ...  # My polars dataframe
pdf = pandas.from_dataframe(pl_df.cols(...).slice_rows(...))
```

kkraus14:
Would be good to have others chime in here given this interchange protocol is already being adopted where we probably don't want to introduce something and later decide to change / remove it.

MarcoGorelli:
It's what plotly already does to not have to convert the entire dataframe

MarcoGorelli:
Any updates here please?

This is the only thing I plan to try adding to the interchange protocol, promised

I think of the interchange protocol as being useful to converting between libraries and doing some preselection in a standardised way:
- select columns (currently possible)
- select rows (not possible)

cc @rgommers @jorisvandenbossche

MarcoGorelli:
gentle ping

(would really like to get this in for pandas 3.0 tbh)

anmyachev:
> I think of the interchange protocol as being useful to converting between libraries and doing some preselection in a standardised way:

> select columns (currently possible)
> select rows (not possible)

The ability to select subset rows in addition to selecting columns seems harmonious.

Implementation in Modin should not be a problem.

+1

MarcoGorelli:
Any updates please?

MarcoGorelli:
closing due to lack of interest (this PR has been open for 5 months), thanks all for comments

